PR #9, opened 2016-05-27
Update absolute position

### LINKED ISSUE #7: Queue status
Hello @hzeller , 
is it possible, from the actual state of the code, to get the queue status from an external process (for example monitoring)? If the code is not ready yet and if you are interested in this feature, how would you like it to be implemented?

lromor:
closes #7
Implemented M114 command.

Now it should be possible to retrieve the precise actual position in realtime of the machine.
pru_motion_queue should be able to callback in the free time motoroperations::status_listener 
so that for example if we are waiting for a slot to be freed, beagleg can still update the actual position and expose it in some other ways (tbd).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -58,7 +58,7 @@ MAIN_OBJECTS=machine-control.o gcode-print-stats.o gcode2ps.o
 TEST_FRAMEWORK_OBJECTS=gtest-all.o gmock-all.o
 
 TARGETS=../machine-control ../gcode-print-stats gcode2ps
-UNITTEST_BINARIES=gcode-machine-control_test gcode-parser_test config-parser_test string-util_test machine-control-config_test arc-gen_test planner_test
+UNITTEST_BINARIES=gcode-machine-control_test gcode-parser_test config-parser_test string-util_test machine-control-config_test arc-gen_test planner_test pru-motion-queue_test
 
 all : $(TARGETS)
 
--- a/src/motion-queue.h
+++ b/src/motion-queue.h
@@ -73,6 +73,14 @@ struct MotionSegment {
 #endif
 } __attribute__((packed));
 
+// Layout of the status register
+// Assuming atomicity of 32 bit boundaries
+// This 32 bit value will be a copy of the R28 register of the PRU
+struct QueueStatus {
+  uint32_t counter : 24; // remaining number of cycles to be performed
+  uint32_t index : 8;    // represent the executing slot [0 to QUEUE_LEN - 1]
+};
+
 // Low level motion queue operations.
 class MotionQueue {
 public:
@@ -91,6 +99,10 @@ public:
 
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
+
+  // Fill the argument with the current absolute position in loops
+  // for each motor.
+  virtual void GetMotorsLoops(int32_t *absolute_pos_loops) = 0;
 };
 
 // Standard implementation.
@@ -100,14 +112,17 @@ public:
 // called.
 class HardwareMapping;
 struct PRUCommunication;
+struct HistorySegment;
 class PRUMotionQueue : public MotionQueue {
 public:
   PRUMotionQueue(HardwareMapping *hw, PruHardwareInterface *pru);
+  ~PRUMotionQueue();
 
   void Enqueue(MotionSegment *segment);
   void WaitQueueEmpty();
   void MotorEnable(bool on);
   void Shutdown(bool flush_queue);
+  void GetMotorsLoops(int32_t *absolute_pos_loops);
 
 private:
   bool Init();
@@ -117,6 +132,15 @@ private:
 
   volatile struct PRUCommunication *pru_data_;
   unsigned int queue_pos_;
+
+  // Shadow Queue
+  void RegisterHistorySegment(MotionSegment *element);
+  void UpdateAbsolutePosition();
+
+  struct HistorySegment *shadow_queue_;
+  int32_t absolute_pos_loops_[MOTION_MOTOR_COUNT];
+  int32_t segment_partials_[MOTION_MOTOR_COUNT];
+  unsigned int back_shadow_queue_;
 };
 
 
@@ -127,6 +151,7 @@ public:
   void WaitQueueEmpty() {}
   void MotorEnable(bool on) {}
   void Shutdown(bool flush_queue) {}
+  void GetMotorsLoops(int32_t *absolute_pos_loops) {}
 };
 
 #endif  // _BEAGLEG_MOTION_QUEUE_H_
--- a/src/motor-interface-pru.p
+++ b/src/motor-interface-pru.p
@@ -28,6 +28,7 @@
 #define CONST_PRUDRAM	   C24
 
 #define QUEUE_ELEMENT_SIZE (SIZE(QueueHeader) + SIZE(TravelParameters))
+#define STATUS_VALUE (QUEUE_LEN * QUEUE_ELEMENT_SIZE)
 
 #define PARAM_START r7
 #define PARAM_END  r19
@@ -167,13 +168,26 @@ calc_decel:
 DONE_CALCULATE_DELAY:
 .endm
 
+;;; This macro decrease the counter that holds the overall number of loops left
+;;; to be performed and then it push it in the PRU DRAM status register.
+.macro UpdateQueueStatus
+	;; Decrease the step counter
+	SUB r28, r28, 1 ; status_loops--
+	;; Push in DRAM
+	MOV r0, STATUS_VALUE
+	SBCO r28, CONST_PRUDRAM, r0, 4
+.endm
+
 INIT:
 	;; Clear STANDBY_INIT in SYSCFG register.
 	LBCO r0, C4, 4, 4
 	CLR r0, r0, 4
 	SBCO r0, C4, 4, 4
 
-	MOV r2, 0		; Queue address in PRU memory
+	MOV r2, 0   ; Queue address in PRU memory
+	MOV r28, 0  ; Status register in PRU memory,
+	            ; r28.b3 for current queue position,
+	            ; bottom three for the remaining steps of the current slot.
 QUEUE_READ:
 	;;
 	;; Read next element from ring-buffer
@@ -202,14 +216,30 @@ QUEUE_READ:
 	ZERO &mstate, SIZE(mstate)	; clear the motor states
 	ZERO &r3, 4			; initialize delay calculation state register.
 
+	;; STATUS REGISTER
+	;; ! We are assuming that writing the 4 bytes status register is atomic
+	;; and we guarantee that the bottom three bytes are all zero so we just need
+	;; to sum up the 3 loop counters. The upper bound of this sum will always be
+	;; less than 2^18. At each loop executed this counter is decreased of one unit.
+	ADD r28, r28, travel_params.loops_accel
+	ADD r28, r28, travel_params.loops_travel
+	ADD r28, r28, travel_params.loops_decel
+	;; Add one more because the first delay is pre-calculated
+	ADD r28, r28, 1
+
+	;; We store a 32-bit word at the end of the queue chunk of memory
+	MOV r0, STATUS_VALUE
+	SBCO r28, CONST_PRUDRAM, r0, 4
+
 	;; Registers
 	;; r0, r1 free for calculation
 	;; r2 = queue pos
 	;; r3 = state for CalculateDelay
-	;; scratch:      r4..r6
-	;; parameter:    r7..r19
-	;; motor-state: r20..r27
-	;; call/ret:    r30
+	;; scratch:           r4..r6
+	;; parameter:         r7..r19
+	;; motor-state:       r20..r27
+	;; status-variable:   r28
+	;; call/ret:          r30
 STEP_GEN:
 	;;
 	;; Generate motion profile configured by TravelParameters
@@ -230,6 +260,7 @@ STEP_GEN:
 	CALL SetSteps
 
 	CalculateDelay r1, travel_params, r3, r5, r6
+	UpdateQueueStatus
 	QBEQ DONE_STEP_GEN, r1, 0       ; special value 0: all steps consumed.
 STEP_DELAY:				; Create time delay between steps.
 	SUB r1, r1, 1                   ; two cycles per loop.
@@ -245,9 +276,11 @@ DONE_STEP_GEN:
 
 	;; Next position in ring buffer
 	ADD r2, r2, QUEUE_ELEMENT_SIZE
+	ADD r28.b3, r28.b3, 1                  ; add + 1 to the MSB byte
 	MOV r1, QUEUE_LEN * QUEUE_ELEMENT_SIZE ; end-of-queue
 	QBLT QUEUE_READ, r1, r2
 	ZERO &r2, 4
+	ZERO &r28, 4
 	JMP QUEUE_READ
 
 FINISH:
--- a/src/pru-motion-queue.cc
+++ b/src/pru-motion-queue.cc
@@ -26,6 +26,7 @@
 #include <errno.h>
 #include <stdio.h>
 #include <strings.h>
+#include <stdlib.h>
 
 #include "generic-gpio.h"
 #include "pwm-timer.h"
@@ -41,7 +42,8 @@
 // an endswitch fires.
 struct PRUCommunication {
   volatile struct MotionSegment ring_buffer[QUEUE_LEN];
-};
+  volatile struct QueueStatus status;
+} __attribute__((packed));
 
 #ifdef DEBUG_QUEUE
 static void DumpMotionSegment(volatile const struct MotionSegment *e,
@@ -79,6 +81,80 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
 }
 #endif
 
+struct HistorySegment {
+  uint32_t total_loops;
+  uint32_t fractions[MOTION_MOTOR_COUNT];
+  uint8_t direction_bits;
+};
+
+void PRUMotionQueue::RegisterHistorySegment(MotionSegment *element) {
+  const unsigned int last_insert_position = (queue_pos_ - 1) % QUEUE_LEN;
+  struct HistorySegment *new_slot = &shadow_queue_[last_insert_position];
+
+  bzero(new_slot, sizeof(struct HistorySegment));
+
+  new_slot->total_loops += element->loops_accel;
+  new_slot->total_loops += element->loops_travel;
+  new_slot->total_loops += element->loops_decel;
+
+  new_slot->direction_bits = element->direction_bits;
+
+  for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
+    new_slot->fractions[i] = element->fractions[i];
+  }
+}
+
+static void fractions_to_loops(const struct HistorySegment &segment,
+                               int32_t *motors_loops) {
+  const uint64_t max_fraction = 0xFFFFFFFF;
+  for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
+    const int sign = (segment.direction_bits >> i) & 1 ? -1 : 1;
+    uint64_t loops = (uint64_t) segment.fractions[i] * (uint64_t) segment.total_loops;
+    loops /= max_fraction;
+    motors_loops[i] = sign * loops;
+  }
+}
+
+void PRUMotionQueue::UpdateAbsolutePosition() {
+  const struct QueueStatus status = *(struct QueueStatus*) &pru_data_->status;
+  struct HistorySegment current = shadow_queue_[status.index];
+  const int shadow_queue_delay
+    = (status.index + QUEUE_LEN - back_shadow_queue_) % QUEUE_LEN;
+  int32_t motors_loops[MOTION_MOTOR_COUNT];
+
+  for (int i = 0; i < shadow_queue_delay; ++i) {
+    fractions_to_loops(shadow_queue_[back_shadow_queue_], motors_loops);
+    for (int j = 0; j < MOTION_MOTOR_COUNT; ++j) {
+      absolute_pos_loops_[j] += motors_loops[j];
+    }
+    shadow_queue_[back_shadow_queue_].total_loops = 0;
+    back_shadow_queue_ = (back_shadow_queue_ + 1) % QUEUE_LEN;
+  }
+
+  // Now update the current executing slot.
+  if (current.total_loops) {
+    uint64_t delta;
+    fractions_to_loops(current, motors_loops);
+    for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
+      absolute_pos_loops_[i] -= segment_partials_[i];
+
+      delta = status.counter * abs(motors_loops[i]);
+      delta /= current.total_loops;
+
+      segment_partials_[i]
+        = motors_loops[i] < 0 ? motors_loops[i] + delta
+                              : motors_loops[i] - delta;
+
+      absolute_pos_loops_[i] += segment_partials_[i];
+    }
+  }
+}
+
+void PRUMotionQueue::GetMotorsLoops(int32_t *absolute_pos_loops) {
+  UpdateAbsolutePosition();
+  memcpy(absolute_pos_loops, absolute_pos_loops_, sizeof(absolute_pos_loops_));
+}
+
 // Stop gap for compiler attempting to be overly clever when copying between
 // host and PRU memory.
 static void unaligned_memcpy(volatile void *dest, const void *src, size_t size) {
@@ -107,6 +183,12 @@ void PRUMotionQueue::Enqueue(MotionSegment *element) {
 
   // Fully initialized. Tell busy-waiting PRU by flipping the state.
   queue_element->state = state_to_send;
+
+  // Update the HistoryQueue before registering a new slot.
+  UpdateAbsolutePosition();
+
+  // Register the last inserted motion segment in the shadow queue.
+  RegisterHistorySegment(element);
 #ifdef DEBUG_QUEUE
   DumpMotionSegment(queue_element, pru_data_);
 #endif
@@ -134,6 +216,8 @@ void PRUMotionQueue::Shutdown(bool flush_queue) {
   MotorEnable(false);
 }
 
+PRUMotionQueue::~PRUMotionQueue() { delete shadow_queue_; }
+
 PRUMotionQueue::PRUMotionQueue(HardwareMapping *hw, PruHardwareInterface *pru)
                                : hardware_mapping_(hw),
                                  pru_interface_(pru) {
@@ -141,6 +225,12 @@ PRUMotionQueue::PRUMotionQueue(HardwareMapping *hw, PruHardwareInterface *pru)
   // For now, we just assert-fail here, if things fail.
   // Typically hardware-doomed event anyway.
   assert(success);
+
+  // Initialize Shadow queue
+  shadow_queue_ = new HistorySegment[QUEUE_LEN];
+  bzero(absolute_pos_loops_, sizeof(absolute_pos_loops_));
+  bzero(segment_partials_, sizeof(segment_partials_));
+  back_shadow_queue_ = 0;
 }
 
 bool PRUMotionQueue::Init() {
--- a/src/pru-motion-queue_test.cc
+++ b/src/pru-motion-queue_test.cc
@@ -11,6 +11,7 @@
 #include <string.h>
 
 #include <gtest/gtest.h>
+#include <gmock/gmock.h>
 
 #include "logging.h"
 #include "motor-operations.h"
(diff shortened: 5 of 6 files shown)


sample.config
```diff
@@ -9,8 +9,8 @@
 # Homing order. For 3D printers, you'd probably home Z last, while CNC machines
 # typically home Z first.
 home-order        = XYZ
-require-homing    = yes  # Require homing (G28) is executed before first move.
-range-check       = yes  # Check that axes are within range. Dangerous if no.
+require-homing    = no  # Require homing (G28) is executed before first move.
```
hzeller: Can  you leave out these unrelated changed from the pull request ?

leonardoromor: Sure, sorry, forgot about this

bigguiness:
On Friday, May 27, 2016 4:52 PM leonardoromor wrote:
> Implemented M114 command.
>
> Now it should be possible to retrieve the precise actual position
> in realtime of the machine.
> pru_motion_queue should be able to callback in the free time
> motoroperations::status_listener so that for example if we are
> waiting for a slot to be freed, beagleg can still update the actual
> position and expose it in some other ways (tbd).

Hello,

Just my two cents... Henner may feel differently.

I think you should rebase this series a bit.

There are a number of commits that add files that are then removed
later. This just clutters the git history and makes it harder to debug
if something breaks later.

Each commit should either "fix" something or add a new feature. Having
commits that just introduce your development steps make it harder to
follow/review the series. The code should be able to compile and run
properly after each commit. If it doesn't the commit needs to be rebased
so that things are not "broken".

The "captians_diary.txt" shouldn't be included in the commits. Those are
your development notes and have nothing to do with the repo code.

Don't change the BEAGLEG_HARDWARE_TARGET in the Makefiles. You can
specify  a different target in the make command line.

Don't commit build files to the repo (build/pip-delete-this-directory.txt).

I would like to review/test this properly but with all the clutter it's a bit
hard to follow.

Over all I think the series is a good idea but currently it's a bit muddled.
Like I said Henner may feel differently.

Regards,
Hartley

leonardoromor:
@bigguiness  Thank you a lot, sorry for the clumsy series, I will try to fix everything you said. I found out just now that usually people cleanup the git history before pull requesting, I feel very noob XD.

hzeller:
Thanks for the cleanup. I agree with Hartley in all the points, and you seemed to have cleaned it up nicely (have not looked at the code yet). While working on stuff, there is a lot of work-in-progress, but having the final pull request one coherent submit with the minimal changes to make the particular feature work without changing unrelated things makes the review more smoohtly.

I'll do a first round of comments, and many will initially just be little style nits and later I'll go into the technical details

src/container.h
```diff
@@ -92,4 +92,53 @@ class RingDeque {
   T buffer_[CAPACITY];
 };
 
+//Stupid reimplementation of the deque
```
hzeller: Whenever you use comments in this code, please add a space after the // (also many other places in this pull request).

Regarding this TODO: is this something you intend to address, or is this indeed a class that you want to keep, or want to use the RingDeque ? In any case, it might need an update in the comment.

src/gcode-machine-control.cc
```diff
@@ -1066,6 +1070,8 @@ void GCodeMachineControl::Impl::input_idle(bool is_first) {
     set_fanspeed(0);
     next_auto_disable_fan_ = -1;
   }
+  //If is idle, we want to update and print ouy the machine position?
```
hzeller: Is this TODO/question resolvable ?

src/gcode-machine-control.cc
```diff
@@ -530,13 +529,21 @@ void GCodeMachineControl::Impl::set_output_flags(HardwareMapping::LogicOutput ou
...
 void GCodeMachineControl::Impl::get_current_position() {
-  if (planning_buffer_.size() > 0) {
-    const struct AxisTarget *current = planning_buffer_[0];
-    const int *mpos = current->position_steps;
-    const float x = 1.0f * mpos[AXIS_X] / cfg_.steps_per_mm[AXIS_X];
-    const float y = 1.0f * mpos[AXIS_Y] / cfg_.steps_per_mm[AXIS_Y];
-    const float z = 1.0f * mpos[AXIS_Z] / cfg_.steps_per_mm[AXIS_Z];
-    const float e = 1.0f * mpos[AXIS_E] / cfg_.steps_per_mm[AXIS_E];
+    int i, motor;
+    const MotorCoordSystem
```
hzeller: I think it is easier to read if you break things like this
   const Type variable
           =  things_to_assign();

src/gcode-machine-control.cc
```diff
@@ -530,13 +529,21 @@ void GCodeMachineControl::Impl::set_output_flags(HardwareMapping::LogicOutput ou
...
+        &motor_position_steps_ = motor_ops_->GetCurrentMotorPos();
```
hzeller: The underscore should only be used for instance variables, not local variables.

You are assigning this to a const reference here. This will work the way c++ guarantees temporary references to exist in the context, but it is probably not what you meant (but instead, meant to copy the coordinates).

Speaking of which, is it not possible hat GetCurrentMotorPos() returns a reference to the current position ? I guess I find out later in the review.

src/gcode-machine-control.cc
```diff
@@ -530,13 +529,21 @@ void GCodeMachineControl::Impl::set_output_flags(HardwareMapping::LogicOutput ou
...
+    FixedArray<int,GCODE_NUM_AXES> axis_position_steps_;
```
hzeller: Same thing here: no underscore in local variables (ok, I'll not mention it again).

src/hardware-mapping.cc
```diff
@@ -97,6 +97,21 @@ bool HardwareMapping::AddMotorMapping(LogicAxis axis, int motor,
   return true;
 }
 
+int  HardwareMapping::GetMotorFromAxis(int axis) {
```
hzeller: are these two spaces after the 'int' ?

src/hardware-mapping.h
```diff
@@ -133,6 +133,11 @@ class HardwareMapping {
   // A value of 0 for 'motor' is accepted but does not connect it to anything.
   bool AddMotorMapping(LogicAxis axis, int motor, bool mirrored);
 
+  // Return the index of the first motor associated to an axis
+  // if the value is negative, the motor has not been mapped.
+  // i is the axis index
+  int GetMotorFromAxis(int i);
```
hzeller: Shouldn't the parameter be of LogicAxis type ? Also, I'd probably name it 'axis'.

Maybe 'from' should be 'for' ? After all, we are looking to get the motor that is for a particular axis.

Given that the function only returns a partial result, I'd call it GetFirstMotorForAxis().

src/hardware-mapping.cc
```diff
@@ -97,6 +97,21 @@ bool HardwareMapping::AddMotorMapping(LogicAxis axis, int motor,
...
+    int i = 0;
+    // Map back the steps to the correct logical axis.
```
hzeller: This comment doesn't make sense.

src/hardware-mapping.cc
```diff
@@ -97,6 +97,21 @@ bool HardwareMapping::AddMotorMapping(LogicAxis axis, int motor,
...
+    uint8_t temp;
```
hzeller: You can do the assignment in one step to save vertical space
  uing8_t temp = axis_to_driver_[axis];

src/hardware-mapping.cc
```diff
@@ -97,6 +97,21 @@ bool HardwareMapping::AddMotorMapping(LogicAxis axis, int motor,
...
+    // Get the axis to driver bitmap
+    temp = axis_to_driver_[axis];
+    // while the first bit is not 1, iterate.
+    while ( ((temp & 0x01) != 0x01) && i < NUM_MOTORS) {
```
hzeller: Maybe rename i as m for 'motor'.

There is an extra space after the opening parenthesis of after while, but not space at the end of the expression. In general: don't add unnecessary space after the opening parenthesis.

src/motion-queue.h
```diff
@@ -72,9 +72,23 @@ struct MotionSegment {
 #endif
 } __attribute__((packed));
 
+//Define a struct to easily access the status register
+struct QueueStatus{
```
hzeller: Add a space between QueueStatus and {

src/motion-queue.h
```diff
@@ -72,9 +72,23 @@ struct MotionSegment {
 #endif
 } __attribute__((packed));
 
+//Define a struct to easily access the status register
```
hzeller: Don't write what you do ('Define a struct'), but rather what it is. "Layout of the status register" for instance.

src/motion-queue.h
```diff
@@ -72,9 +72,23 @@ struct MotionSegment {
...
+struct QueueStatus{
+    uint32_t counter : 24;
+    uint32_t index : 8;
+}__attribute__((packed));
```
hzeller: Be consistent: above, we have a space between } and the __attribute__

src/motion-queue.h
```diff
@@ -72,9 +72,23 @@ struct MotionSegment {
...
+
+
 // Low level motion queue operations.
 class MotionQueue {
 public:
+    class EventReceiver {
+    public:
+      virtual void update_absolute_position() = 0; //callback to update the absolute position
```
hzeller: This does not really explain what is happening. The name implies that this receives some kind of Event. And apparently that is update_absolute_position(). But wouldn't we expect with then that this contains the new information in the parameter or something ?

Is 'EventReceiver' a good name ? Can we be more specific (PositionChangeCalback or something ?). (Haven't looked at the rest of the code yet, so I might mis-understand. But I think things should be immediately understandable for someone reading the header file).

src/motion-queue.h
```diff
@@ -90,6 +104,11 @@ class MotionQueue {
 
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
+
+  //Return the pointer to the status register.
+  virtual const struct QueueStatus get_status() = 0;
```
hzeller: I prefer to have 'cheap' methods be lower-case, everything else Uppercase. GetStatus() maybe ?

src/motion-queue.h
```diff
@@ -101,15 +120,13 @@ class HardwareMapping;
...
 public:
   PRUMotionQueue(HardwareMapping *hw);
-
   void Enqueue(MotionSegment *segment);
   void WaitQueueEmpty();
   void MotorEnable(bool on);
   void Shutdown(bool flush_queue);
-
+  const struct QueueStatus get_status();
 private:
```
hzeller: Please leave a newline between a function declaration and the next 'private'.

src/motor-interface-constants.h
```diff
@@ -95,6 +95,12 @@
 #define PIN_P8_19  (GPIO_0_BASE | 22)  // EHRPWM2A
 #define PIN_P8_26  (GPIO_1_BASE | 29)
 
+#define PIN_P8_43  (GPIO_2_BASE | 8) //Check http://www.ofitselfso.com/BeagleNotes/pinmux.pdf, select the GPIO device and the number
```
hzeller: Is the URL a relevant information that we need in particular for PIN_P8_43 ? I think the general documentation about GPIO things should be at a different place.

src/motor-interface-pru.p
```diff
@@ -28,6 +28,7 @@
 #define CONST_PRUDRAM	   C24
 
 #define QUEUE_ELEMENT_SIZE (SIZE(QueueHeader) + SIZE(TravelParameters))
+#define QUEUE_END QUEUE_LEN * QUEUE_ELEMENT_SIZE
```
hzeller: Always put parenthesis around expressions you declare in defines, you never know in which operator precendence context it is used.

src/motor-interface-pru.p
```diff
@@ -165,6 +167,15 @@ calc_decel:
 	SUB output_reg, output_reg, (IDIV_MACRO_CYCLE_COUNT + 11) / 2
 
 DONE_CALCULATE_DELAY:
+
+.endm
+
+.macro UpdateQueueStatus
```
hzeller: A little description of what this is doing is probably a good idea.

src/motor-interface-pru.p
```diff
@@ -174,6 +185,9 @@ INIT:
 	SBCO r0, C4, 4, 4
 
 	MOV r2, 0		; Queue address in PRU memory
+	MOV r28, 0		; Status register in PRU memory,
+	 				; top byte for current queue position,
```
hzeller: Can you align the ';' with the line before ?
(or does it just look off in the diff because of some TAB ?)

src/motor-interface-pru.p
```diff
@@ -245,9 +276,11 @@ DONE_STEP_GEN:
 
 	;; Next position in ring buffer
 	ADD r2, r2, QUEUE_ELEMENT_SIZE
+	ADD r28.b3, r28.b3, 1 ; add + 1 to the MSB byte
```
hzeller: I'd probably move the ';' comment more out for readability; so more space before the ';'.

src/motor-operations.cc
```diff
@@ -84,19 +85,97 @@ static char test_acceleration_ok(float acceleration) {
 }
 #endif
 
+struct history_slot{
```
hzeller: Add a space between type-name and open curly brace.

Regarding type names: I prefer upper case names in the context of this project, so more like HistorySlot.

src/motor-operations.cc
```diff
@@ -84,19 +85,97 @@ static char test_acceleration_ok(float acceleration) {
...
+typedef FixedArray<unsigned,BEAGLEG_NUM_MOTORS> SlotPartial;
+
+
+// This class is delegated to handle the update of the absolute
+// position in steps of each motor and it should be callbacked whenever this operations
+// has to be performed.
+class MotionQueueMotorOperations::StatusListener : public MotionQueue::EventReceiver {
+public:
+    StatusListener(MotionQueue *backend, MotorCoordSystem *current_position_steps)
+                        : backend_(backend), current_position_steps_(current_position_steps) {};
```
hzeller: Try to stay within 80 characters.
There is an unnecessary semicolon at the end of that line.

src/motor-operations.cc
```diff
@@ -84,19 +85,97 @@ static char test_acceleration_ok(float acceleration) {
...
+    virtual void update_absolute_position();
+    void push_in_history(const LinearSegmentSteps &param, const int defining_axis_steps) {
+        struct history_slot *just_uploaded = execution_history.append();
+        for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
+            just_uploaded->steps[i] = param.steps[i];
+        }
+        just_uploaded->max_axis_steps = defining_axis_steps;
+    }
+private:
```
hzeller: (newline above 'private:')

src/motor-operations_test.cc
```diff
@@ -0,0 +1,126 @@
+/* -*- mode: c++; c-basic-offset: 2; indent-tabs-mode: nil; -*-
+ * Test for gcode machine control.
+ *
+ * We simulate inputs coming gcode-machine-control
+ * we push them into a dummy pru queue, and check if the history queue correctly evaluates
```
hzeller: Try to stay within 80 characters.

src/motor-operations.cc
```diff
@@ -84,19 +85,97 @@ static char test_acceleration_ok(float acceleration) {
...
+    MotionQueue *const backend_;
+
+    SlotHistory execution_history;
+    SlotPartial last_partial;
+    MotorCoordSystem *const current_position_steps_;
+};
+
+void MotionQueueMotorOperations::StatusListener::update_absolute_position() {
+    int i;
```
hzeller: Declare the variable as close as possible where it is used. Ideally within the for loop directly.

src/motor-operations.cc
```diff
@@ -84,19 +85,97 @@ static char test_acceleration_ok(float acceleration) {
...
+    status_listener_->update_absolute_position();
+    return current_position_steps;
+}
+
 void MotionQueueMotorOperations::EnqueueInternal(const LinearSegmentSteps &param,
                                                  int defining_axis_steps) {
   struct MotionSegment new_element = {0};
   new_element.direction_bits = 0;
-
+  int i;
```
hzeller: What was the reason for this change ? You removed the 'int' from the for-loop declaration and put it here. Don't do that unless absolutely necessary to avoid programming errors.

src/motor-operations.h
```diff
@@ -64,22 +69,30 @@ class MotorOperations {  // Rename SegmentQueue ?
 
   // Wait, until all elements in the ring-buffer are consumed.
   virtual void WaitQueueEmpty() = 0;
+
+  // Get a reference of a FixedArray containing the absolute motors
```
hzeller: Here you talk about getting a reference, but instead, you return a value. Is the comment outdated or did you indeed meant to return a reference ?

src/motor-operations.h
```diff
@@ -64,22 +69,30 @@ class MotorOperations {  // Rename SegmentQueue ?
...
+  // positions in steps.
+  virtual const MotorCoordSystem GetCurrentMotorPos() = 0;
 };
 
 class MotionQueueMotorOperations : public MotorOperations {
 public:
   // Initialize motor operations, sending planned results into the motion backend.
-  MotionQueueMotorOperations(MotionQueue *backend) : backend_(backend) {}
+  MotionQueueMotorOperations(MotionQueue *backend);
+  ~MotionQueueMotorOperations();
```
hzeller: Even though not needed, it is always a good idea to prepend this with 'virtual' for the user to read (once we decide to use c++11 flags, we can have 'override' instead at the end, but we don't have that at this point).

src/motor-operations.h
```diff
@@ -64,22 +69,30 @@ class MotorOperations {  // Rename SegmentQueue ?
...
   virtual void MotorEnable(bool on);
   virtual void WaitQueueEmpty();
-
+  virtual const MotorCoordSystem GetCurrentMotorPos();
 private:
   void EnqueueInternal(const LinearSegmentSteps &param,
                        int defining_axis_steps);
-
   MotionQueue *backend_;
+private:
```
hzeller: superfluous private ?

src/pru-motion-queue.cc
```diff
@@ -107,15 +111,23 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e) {
 }
 #endif
 
+const struct QueueStatus PRUMotionQueue::get_status(){
+    //Snapshot the status register
+    struct QueueStatus status_register;
+    memcpy(&status_register, (struct QueueStatus *)&pru_data_->status, sizeof(struct QueueStatus));
```
hzeller: Given that you anyway memcpy() things, mabye GetStatus() should actually get a QueueStatus* paramter that is filled in.

hzeller:
Added a first round of comments; not yet looked too much into detail of the technical things.

There are some strategic points in which you should probably explain the feedback between PRU and host (the StatusRegister thing). And the assumptions you make (e.g. that a 32bit operation is atomic, hence the 24/8 bit split), so that it is easier for future readers to understand.

hzeller:
(Ah, and don't worry, it looks like a lot, but I am also a pedantic code reviewer .. as you have seen in FlaschenTaschen :) ... most of the comments are minor)

src/container.h
bigguiness: Your RingQueue class _looks_ like it does the same thing as the RingDeque class. You just added a queue_size_ member instead of calculating the value based on the back_pos_/front_pos_ (read_pos_/write_pos_ in RingDeque) and added accessors for the back_pos_/fron_pos_t members. It would be cleaner to use the existing class and add the accessors to that class (if they are actually needed).

src/gcode-machine-control.cc
```diff
@@ -62,7 +62,6 @@
 // used in planning along the path.
 struct AxisTarget {
   int position_steps[GCODE_NUM_AXES];  // Absolute position at end of segment. In steps.
-
   // Derived values
```
bigguiness: Unrelated whitespace removed.

src/gcode-machine-control.cc
```diff
@@ -530,13 +529,21 @@ void GCodeMachineControl::Impl::set_output_flags(HardwareMapping::LogicOutput ou
...
 
 void GCodeMachineControl::Impl::get_current_position() {
-  if (planning_buffer_.size() > 0) {
-    const struct AxisTarget *current = planning_buffer_[0];
-    const int *mpos = current->position_steps;
-    const float x = 1.0f * mpos[AXIS_X] / cfg_.steps_per_mm[AXIS_X];
-    const float y = 1.0f * mpos[AXIS_Y] / cfg_.steps_per_mm[AXIS_Y];
-    const float z = 1.0f * mpos[AXIS_Z] / cfg_.steps_per_mm[AXIS_Z];
-    const float e = 1.0f * mpos[AXIS_E] / cfg_.steps_per_mm[AXIS_E];
+    int i, motor;
```
bigguiness: The whitespace indent is off in this function.

src/hardware-mapping.cc
```diff
@@ -97,6 +97,21 @@ bool HardwareMapping::AddMotorMapping(LogicAxis axis, int motor,
   return true;
 }
 
+int  HardwareMapping::GetMotorFromAxis(int axis) {
+    int i = 0;
```
bigguiness: The whitespace indent looks wrong. Henner's CodingStyle is 2 space indents.

src/motor-interface-constants.h
```diff
@@ -95,6 +95,12 @@
...
+#define PIN_P8_44  (GPIO_2_BASE | 9)
+#define PIN_P8_40  (GPIO_2_BASE | 13)
+#define PIN_P8_39  (GPIO_2_BASE | 12)
+#define PIN_P8_28  (GPIO_2_BASE | 24)
```
bigguiness: Please sort these by pin number like the other PIN_P8_* defines.
Also please add a comment about the "standard" pin name like the others (these are all LCD_* pins).

src/Makefile
```diff
@@ -105,7 +105,7 @@ gmock-all.o : $(GMOCK_SOURCE)/src/gmock-all.cc compiler-flags
 	$(CROSS_COMPILE)$(CXX) $(CXXFLAGS) -I$(GMOCK_SOURCE) -c  $< -o $@
 
 clean:
-	rm -rf $(TARGETS) $(MAIN_OBJECTS) $(OBJECTS) $(PRU_BIN) $(UNITTEST_BINARIES) $(OBJECTS:.o=.d) $(MAIN_OBJECTS:.o=.d) $(TEST_FRAMEORK_OBJECTS)
+	rm -rf $(TARGETS) $(MAIN_OBJECTS) $(OBJECTS) $(PRU_BIN) $(UNITTEST_BINARIES) $(OBJECTS:.o=.d) $(MAIN_OBJECTS:.o=.d) $(TEST_FRAMEWORK_OBJECTS)
 
 compiler-flags: FORCE
 	@echo '$(CXX) $(CXXFLAGS)' | cmp -s - $@ || echo '$(CXX) $(CXXFLAGS)' > $@
```
bigguiness: Henner,

You might want to cherry-pick this part. It's just a typo fix and not related to the series.

Hartley

src/gcode-machine-control.cc
leonardoromor: Old useless thoughts

src/gcode-machine-control.cc
leonardoromor: SOLVED

src/hardware-mapping.cc
leonardoromor: (solved)

src/hardware-mapping.cc
leonardoromor: Meh :\ 
(solved)

src/hardware-mapping.cc
leonardoromor: (solved)

src/hardware-mapping.cc
leonardoromor: (solved)

src/motion-queue.h
leonardoromor: (solved)

src/motion-queue.h
leonardoromor: (solved)

src/motion-queue.h
leonardoromor: (solved)

src/motor-interface-constants.h
leonardoromor: (solved)

src/hardware-mapping.cc
leonardoromor: (solved and checking everywhereelse)

src/gcode-machine-control.cc
leonardoromor: (solved)

src/gcode-machine-control.cc
leonardoromor: (solved)

src/motor-operations.h
leonardoromor: (solved)

src/motor-operations.cc
leonardoromor: (solved)

src/motor-operations_test.cc
leonardoromor: (solved)

src/motor-operations.cc
leonardoromor: (solved)

src/motor-operations.cc
leonardoromor: (solved)

src/motor-operations.cc
leonardoromor: (solved)

src/motor-interface-pru.p
leonardoromor: (solved)

src/motor-interface-pru.p
leonardoromor: (solved)

src/motor-interface-pru.p
leonardoromor: (solved)

src/motor-interface-pru.p
leonardoromor: Nice tip. (Solved)

src/motor-interface-constants.h
leonardoromor: (solved)

src/motion-queue.h
leonardoromor: (solved)

src/hardware-mapping.h
leonardoromor: (solved)

src/motor-operations.h
leonardoromor: so I wanted to use the reference initially, but I got stuck by finding a nice way that makes happy also the tests and the mock/dummy classes that inherits from motor_operations where the method isn't implemented yet and had to return a mock MotorCoordSystem without forcing people to add a mock member variable to return.

> Alternative could as well be to give it a pointer to a MotorCoordSystem variable and have the function-call fill in the values.

I guess I will follow this approach.

src/motor-operations.h
leonardoromor: (solved)

leonardoromor:
There's also a new issue, if beagleg is run with dry run, the queue is updated but nothing is executed so at a certain point an assert complains because the PRU is not executing anything. 
Should I replace DummyMotionQueue with the MotionQueue used for the tests that handles a correct GetStatus() behavior?

src/container.h
leonardoromor: I guess the only reason I didn't like it was because you can't append for many times as the capacity of the queue but maybe this is a standard behavior. 
So if I have to mirror the PRU queue that is made of 16 slots I have to make a deque with CAPACITY=17 so I made it work also with this deque.

src/gcode-machine-control.cc
leonardoromor: (solved / look below)

src/gcode-machine-control.cc
leonardoromor: (solved)

src/motion-queue.h
leonardoromor: (solved)

src/pru-motion-queue.cc
leonardoromor: (solved)

leonardoromor:
So,  any other considerations left?

mikikg:
Does this update also cover machine states like in GRBL, for example it would be very useful to have report on machine status, does machine is in RUN, IDLE or ERROR state?
In many cases we will have request to wait until controller finish some movements so we can synchrony executes some other operation initiated from upfront application.

hzeller:
@leonardoromor I am still on my way back from Toorcamp but will have stable internet again tonight.
@mikikg One feature at a time :) But it does sounds like a brittle design of your application if you have to wait for gcdes to be finished to do something. I rather suggest to use the various commands that switch output pins when it is their turn; then you can use these to exactly switch whatever you have to do.

mikikg:
@hzeller Tnx for reply. In my case I need to read some external Inputs and currently that is not supported by BeagleG. I need simple scenario likes, I move working peace "into machine" (hydraulic press) ant that can handle BeagleG but I need to wait in some point when press finish it's job (they will notify me) so I can "pull out" working peace. In that scenario I need synchronisation between motion controller and PLC operations, I can't rely on timing of movement, I need event (or to pull periodically) when motion controller finish it's job.
That was the reason and request for such function in BegleG.

src/Makefile
hzeller: Yep, this would be a good separate change.

src/motor-operations.h
hzeller: What I mean: instead of returning a value, just have the user pass a pointer into the function
```
  GetCurrentMotorPos(MotorCoordSystem *here_I_want_it);
```
... and have the function fill whereever that is pointing to.

Returning a reference or pointer to some internal thing can be problematic as you don't have control over when the caller accesses the value, and if it is something that is changed in parallel; so this is why you returned a value in the first place.

src/motor-operations.h
```diff
@@ -64,22 +69,30 @@ class MotorOperations {  // Rename SegmentQueue ?
 
   // Wait, until all elements in the ring-buffer are consumed.
   virtual void WaitQueueEmpty() = 0;
+
+  // Get a reference of a FixedArray containing the absolute motors
+  // positions in steps.
+  virtual void GetCurrentMotorPos(MotorCoordSystem &current_position_steps) = 0;
```
hzeller: In the context of this code: never use non-const references in parameters. const references are for passing stuff that is not to be changed; if you want to return something, use a pointer, so that it is clear for the caller that this modifies things.

src/motion-queue.h
```diff
@@ -72,9 +75,27 @@ struct MotionSegment {
 #endif
 } __attribute__((packed));
 
+// Layout of the status register
```
hzeller: Maybe mention why you do the 24/8 bit split: assuming atomicity of 32 bit boundaries.

src/motion-queue.h
```diff
@@ -72,9 +75,27 @@ struct MotionSegment {
...
+struct QueueStatus {
+  uint32_t counter : 24;
+  uint32_t index : 8;
+} __attribute__((packed));
```
hzeller: __attribute__((packed)) is probalby not needed for this one. The compiler knows what to do here.
Packed only really is needed if you have multiple fields in a struct which might trigger padding (e.g. a byte next to a int16).

src/motor-interface-constants.h
```diff
@@ -94,6 +94,11 @@
 #define PIN_P8_18  (GPIO_2_BASE |  1)
 #define PIN_P8_19  (GPIO_0_BASE | 22)  // EHRPWM2A
 #define PIN_P8_26  (GPIO_1_BASE | 29)
+#define PIN_P8_28  (GPIO_2_BASE | 24)  // LCD_PCLK
```
hzeller: This looks like it should be a separate change.

hzeller:
After a couple of weeks away busy with other things, looked at the latest change. Looks much better. I have some superficial comments, but I need to take a more closer look tonight regarding the logic

src/motor-interface-constants.h
leonardoromor: right, I will remove it =)

src/motor-operations.h
leonardoromor: Ok, Now I understand better =)

src/motion-queue.h
leonardoromor: Yes, makes sense, sorry.

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
...
+// operations has to be performed.
+class MotionQueueMotorOperations::StatusListener
+                                  : public MotionQueue::PositionChangeCallback {
+public:
+  StatusListener(MotionQueue *backend, MotorCoordSystem *current_position_steps)
+                : backend_(backend),
+                current_position_steps_(current_position_steps) {}
+  virtual void UpdateAbsolutePosition();
+  void PushInHistory(const LinearSegmentSteps &param,
+                       const int defining_axis_steps) {
```
hzeller: this looks like a weird indentation ('const' should be aligned under the 'const' the line above).

src/motion-queue.h
```diff
@@ -72,9 +75,27 @@ struct MotionSegment {
...
+
 // Low level motion queue operations.
 class MotionQueue {
 public:
+  // When the queue is full, beagleg has to wait a slot to be freed from the PRU.
+  // In that free time the execution is blocked waiting for the interrupt.
+  // So it would be impossibile without a thread to ask beagleg to update the
```
hzeller: This is hard to understand.

Who is calling this callback ?

src/motion-queue.h
```diff
@@ -90,6 +111,11 @@ class MotionQueue {
 
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
+
+  // Return the pointer to the status register.
+  virtual void GetStatus(struct QueueStatus *status) = 0;
+
+  PositionChangeCallback *callbacks;
```
hzeller: Never put a member accessible publically. Make a setter if you want to set it.
Describe what it does.

src/motion-queue.h
```diff
@@ -106,21 +132,36 @@ class PRUMotionQueue : public MotionQueue {
...
 };
 
-
 // Queue that does nothing. For testing purposes.
 class DummyMotionQueue : public MotionQueue {
 public:
-  void Enqueue(MotionSegment *segment) {}
+  DummyMotionQueue() : mock_pru_status_() {}
+  void Enqueue(MotionSegment *segment) {
+    if(mock_pru_queue_.size() == QUEUE_LEN) mock_pru_queue_.pop_front();
```
hzeller: always leave a space between the `if` and `(`.

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
 }
 #endif
 
+struct HistorySlot {
```
hzeller: This type needs an explanation

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
...
+  int steps[BEAGLEG_NUM_MOTORS];
+  int max_axis_steps;
+};
+
+typedef RingDeque<struct HistorySlot, QUEUE_LEN + 1> SlotHistory;
```
hzeller: Why does the SlotHistory needs one more element than QUEUE_LEN ?

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
...
+typedef FixedArray<unsigned,BEAGLEG_NUM_MOTORS> SlotPartial;
```
hzeller: Add a space after the comma.

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
...
+
+void MotionQueueMotorOperations::StatusListener::UpdateAbsolutePosition() {
+  int i;
+  struct QueueStatus status_register;
+  backend_->GetStatus(&status_register);
+  if (execution_history_.size() == 0) return;
+  // Pru_queue_size represent the real pru queue size, so we shrink the history
+  // queue until it's exactly the same as the pru one.
+  const unsigned pru_queue_size
+    = ((execution_history_.back_index() - 1) % (QUEUE_LEN + 1)
```
hzeller: This breaks the encapsulation of the execution_history_ ... you are essentially doing things here that you only can do with the knowledge how the execution history is dealt with internally.

Anyway, this part is extremely hard to understand, what is happening here ?

src/motion-queue.h
```diff
@@ -72,9 +75,27 @@ struct MotionSegment {
 #endif
 } __attribute__((packed));
 
+// Layout of the status register
+struct QueueStatus {
+  uint32_t counter : 24;
```
hzeller: These values need an explanation: what is the counter referring to and what is index referring to. Why is it sufficient that we can live with 24 bit here.
Right now, this might be fresh in your mind, but If they are not documented here, then someone else will have a hard time understanding the code ... or you yourself in 3 months after having worked on other parts.

src/motor-interface-pru.p
```diff
@@ -167,13 +168,26 @@ calc_decel:
...
 INIT:
 	;; Clear STANDBY_INIT in SYSCFG register.
 	LBCO r0, C4, 4, 4
 	CLR r0, r0, 4
 	SBCO r0, C4, 4, 4
 
-	MOV r2, 0		; Queue address in PRU memory
+	MOV r2, 0   ; Queue address in PRU memory
+	MOV r28, 0  ; Status register in PRU memory,
+	            ; top byte for current queue position,
```
hzeller: Maybe qualify top-byte as r28.b3

src/motor-interface-pru.p
```diff
@@ -167,13 +168,26 @@ calc_decel:
...
+	            ; bottom for the remaining steps of the current slot.
```
hzeller: .. bottom three

src/motor-interface-pru.p
```diff
@@ -202,14 +216,30 @@ QUEUE_READ:
 	ZERO &mstate, SIZE(mstate)	; clear the motor states
 	ZERO &r3, 4			; initialize delay calculation state register.
 
+	;; STATUS REGISTER
+	;; ! We are assuming that writing the 4 bytes status register is atomic
```
hzeller: Also we assume that the lower 24 bits are all zero when we are here, because we guaraneed that they are zero at the end of the last step.
Given that we just add values here, this might be useful to mention.

src/motor-interface-pru.p
```diff
@@ -202,14 +216,30 @@ QUEUE_READ:
...
+	;; We sum the 3 loop counters inside the last 24 bits of a 4 byte variable
+	;; at the end of the ring buffer. At each step loop executed this counter
+	;; is decreased of one unit
+	ADD r28, r28, travel_params.loops_accel
+	ADD r28, r28, travel_params.loops_travel
+	ADD r28, r28, travel_params.loops_decel
+	;; Add one more because the first delay is pre-calculated
```
hzeller: mmh, can you elaborate ? Is it always one ?
This is not immediately clear to me why that is the case.

src/motor-interface-pru.p
```diff
@@ -28,6 +28,7 @@
 #define CONST_PRUDRAM	   C24
 
 #define QUEUE_ELEMENT_SIZE (SIZE(QueueHeader) + SIZE(TravelParameters))
+#define QUEUE_END (QUEUE_LEN * QUEUE_ELEMENT_SIZE)
```
hzeller: Logically, this would be STATUS_VALUE. It happens to be at the end of the queue, but that is not what logically is there.

src/motor-operations.h
```diff
@@ -21,12 +21,17 @@
...
+#include "gcode-parser.h"
+#include "motor-interface-constants.h"
+
 class MotionQueue;
 
 enum {
   BEAGLEG_NUM_MOTORS = 8
 };
 
+typedef FixedArray<int,BEAGLEG_NUM_MOTORS> MotorCoordSystem;
```
hzeller: Don't forget the space after the comma.

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
...
 
+struct HistorySlot {
+  int steps[BEAGLEG_NUM_MOTORS];
+  int max_axis_steps;
+};
+
+typedef RingDeque<struct HistorySlot, QUEUE_LEN + 1> SlotHistory;
+typedef FixedArray<unsigned,BEAGLEG_NUM_MOTORS> SlotPartial;
+
+// This class is delegated to handle the update of the absolute
```
hzeller: I don't really understand this sentence. 'whenever this operation as to be performed' .. which operation is meant ? Also, it is not clear who is expected to be calling this.

src/container.h
```diff
@@ -86,6 +86,9 @@ class RingDeque {
     read_pos_ = (read_pos_ + 1) % CAPACITY;
   }
 
+  unsigned front_index() const { return read_pos_; }
```
hzeller: Exposing these values breaks encapsulation of this RingDeque; we should avoid it.

You do this to be able to follow the lower level queue and then trying to keep track of the index and attempt to re-create that situation with the deque.

Don't do it that complicated: after all, the queue with the underlying PRU is handled 'manually' anyway: we always know which index we entered something when we stored an element in there. So just let the `MotionQueue::Enqueue(MotionSegment *element)` return an integer which represents the index in which it was stored.

Then things suddenly get simple and understandable: the index in the StatusRegister suddenly makes sense, as it has a 1:1 correlation with the index returned by the Enqueue(); your history queue can just be implemented by simply allocating an array of of QUEUE_SIZE with the elements in there.

src/motor-operations.cc
```diff
@@ -84,6 +85,95 @@ static char test_acceleration_ok(float acceleration) {
...
+      just_uploaded->steps[i] = param.steps[i];
+    }
+    just_uploaded->max_axis_steps = defining_axis_steps;
+  }
+
+private:
+  MotionQueue *const backend_;
+  SlotHistory execution_history_;
+  SlotPartial last_partial;
+  MotorCoordSystem *const current_position_steps_;
```
hzeller: Try to order things: first all the consts, then the rest. So this should come after backend_.

hzeller:
I think in general, it looks pretty cool. I have a number of comments mostly around:

   * Documentation. You need to express your intend when adding types and members. While it might be clear while writing, you need to be able to pass on the idea to someone reading the code (and yourself after not looking at it for a while)
   * Also Documentation and possibly simplification: it is not entirely clear who is calling what up and downstream. Maybe the MotionQueue could even keep track of the counts whenever asked, so that it has not to call up ?
   * Complexity: You go out of your way to use the RingDeque to keep track of the history, but you're breaking abstraction of the RingDeque in the course of that. Also, it is very hard to understand what is going on because of that. Rather, try to just keep track of the history in a separate queue, similarly handled like the PRU motion queue (i.e. 'manual' RingDeque) and keep everything fixed to that. Either by having Enqueu() return an index to where things are and keep a 'shadow' array with the history information (and also, with that index the StatusRegister makes sense, because the index refers to somnething real.). Or even have PRUMotionQueue::Enqueue() allow to store an arbitrary pointer object alongside and make that accessible via the index. Your call :)

src/motion-queue.h
leonardoromor: (Solved)

src/motion-queue.h
leonardoromor: (Solved)

src/motor-interface-pru.p
leonardoromor: (Solved)

src/motor-interface-pru.p
leonardoromor: (Solved)

src/motor-interface-pru.p
leonardoromor: (Solved)

src/motion-queue.h
leonardoromor: (Solved)

hzeller:
Are there updates here ?

leonardoromor:
Not yet, I hope to have some time this weekend to work on it

src/pru-hardware-interface.h
```diff
@@ -0,0 +1,60 @@
...
+public:
+  virtual ~PruHardwareInterface() {}
+
+  // Initialize the hardware, load the binary, interrupts, etc...
+  virtual int Init() = 0;
+
+  // Retrieve the pointer of the pru mapping and initialize the memory.
+  virtual int Allocate(void **pru_mmap, const size_t size) = 0;
+
+  // Enable the PRU
```
hzeller: I would call this StartExecution(). Execute() sounds like it is finished once the method returns.
The comment is a little misleading; it should be more along the lines of 'start predetermined program' or something.

src/pru-hardware-interface.h
```diff
@@ -0,0 +1,60 @@
...
+  virtual int Execute() = 0;
+
+  // Wait for a beagleg-mapped event
+  virtual int WaitEvent(const unsigned int eventnum) = 0;
```
hzeller: At this point: do we need more than one event type ? If not, I'd just leave this out and leave it an implementation detail of the implementor how that is done internally.

Because otherwise, the pru-motion-queue needs to include the PRU headers to know special symbols such as PRU_EVTOUT_0 ... which  means we actually lost the abstraction.

src/pru-hardware-interface.h
```diff
@@ -0,0 +1,60 @@
...
+// Pru hardware controls
+class PruHardwareInterface {
+public:
+  virtual ~PruHardwareInterface() {}
+
+  // Initialize the hardware, load the binary, interrupts, etc...
+  virtual int Init() = 0;
+
+  // Retrieve the pointer of the pru mapping and initialize the memory.
+  virtual int Allocate(void **pru_mmap, const size_t size) = 0;
```
hzeller: Maybe we should call that AllocateSharedMem() ? Because in the abstract way, that is what it is.

What is the return value ? Description is needed.

src/pru-hardware-interface.h
```diff
@@ -0,0 +1,60 @@
...
+
+#define SEGMENT_EXECUTED PRU_EVTOUT0
+
+// Pru hardware controls
+class PruHardwareInterface {
+public:
+  virtual ~PruHardwareInterface() {}
+
+  // Initialize the hardware, load the binary, interrupts, etc...
+  virtual int Init() = 0;
```
hzeller: What is the return value ? I guess we only really can signal success/failure here, so it should probably be bool.
(early code in BeagleG was using more C-convention, so that was typically an integer with 0 = success and other values = failure. If there is really only one failure type, then this 'inverse logic' is pretty confusing).

src/pru-hardware-interface.h
```diff
@@ -0,0 +1,60 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with BeagleG.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef BEAGLEG_PRU_HARDWARE_INTERFACE_
+#define BEAGLEG_PRU_HARDWARE_INTERFACE_
+
+#include <cstddef>
+#include <pruss_intc_mapping.h>
```
hzeller: In this abstraction, this PRU specific include should probably not be here.

Think of it this way: everything should be compilable with some other implementation without even the am335x package in the source path.

hzeller:
Pulling out the PRU code looks very good, pretty much exactly how I'd done it (of course, I still have some comments :) ).
Can you make this a separate pull-request, so that we can integrate that already ?

src/motion-queue.h
```diff
@@ -111,6 +113,17 @@ class PRUMotionQueue : public MotionQueue {
   int Init();
 
   HardwareMapping *const hardware_mapping_;
+  PruHardwareInterface *const pru_interface_;
+
+  // The communication with the PRU. We memory map the static RAM in the PRU
+  // and write stuff into it from here. Mostly this is a ring-buffer with
+  // commands to execute, but also configuration data, such as what to do when
+  // an endswitch fires.
+  struct PRUCommunication {
```
leonardoromor: (Solved)

src/pru-hardware-interface.h
leonardoromor: (Solved)

src/pru-hardware-interface.h
leonardoromor: (Solved)

src/pru-hardware-interface.h
leonardoromor: (Solved)

src/pru-hardware-interface.h
leonardoromor: (Solved): All the methods will return bool except WaitEvent that will return the number of times the event will be triggered.

src/pru-hardware-interface.h
leonardoromor: (Solved)

src/motor-interface-pru.p
leonardoromor: (Solved): QUEUE_END changed into STATUS_VALUE

PR #10, opened 2016-06-02
Accel speed clamping

leonardoromor:
Added clamping function for both speeds and accelerations limits.

Notes:

1.  I think that the user should be notified via socket or logs that because of the settings the speed is impossible to reach so the max has been set.
2. The feedrate should always represent a speed during an euclidean displacement. What happens if we just use the extruder and all the other axis are in stop? The feedrate in that case represent the flowrate/mm/s rpm of the tool. Reading around the correct way of setting the spindle speed or flowrate should be done using the gcode S command. Do you agree?


hzeller:
Merged the test manually in https://github.com/hzeller/beagleg/commit/5294ce90bb9a2cd03a42d813cd0505561c38545b which works with the code in Pull request #17 .
Is there more to pull out of here, or are we ready to close it ?

bigguiness:
Should this be closed?

hzeller:
Yes, this can be closed.

PR #12, opened 2016-06-21
first alignment test to solve BusError, compile and works on kernel 4.4 bone rt

leonardoromor:
i'm not sure if it may produce undefined behaviors better to check.


leonardoromor:
As we feared this patch at some point may produce bad behaviors, with the deferenced copy it may override part of the next pru slot. Do not use this patch.

PR #15, opened 2016-07-28
Pru interface separation

leonardoromor:


src/pru-hardware-interface.h
```diff
@@ -0,0 +1,57 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with BeagleG.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef BEAGLEG_PRU_HARDWARE_INTERFACE_
+#define BEAGLEG_PRU_HARDWARE_INTERFACE_
+
+#include <cstddef>
+
+// Target PRU
+#define PRU_NUM 0
```
hzeller: Is this needed ?

src/pru-motion-queue.cc
```diff
@@ -170,24 +147,17 @@ PRUMotionQueue::PRUMotionQueue(HardwareMapping *hw) : hardware_mapping_(hw) {
 
 int PRUMotionQueue::Init() {
   MotorEnable(false);  // motors off initially.
+  bool status = pru_interface_->Init();
+  if (status) return status;
```
hzeller: Shouldnt' this be if (!status) ?

src/pru-hardware-interface.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Pru hardware controls
+class PruHardwareInterface {
+public:
+  virtual ~PruHardwareInterface() {}
+
+  // Initialize the hardware, interrupts, etc...
```
hzeller: Maybe describe return value: return 'true' on success.

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+#include <strings.h>
+#include <string.h>
+
+#include "logging.h"
+
+// Generated PRU code from motor-interface-pru.p
+#include "motor-interface-pru_bin.h"
+
+#if PRU_NUM==0
+#define PRU_DATARAM PRUSS0_PRU0_DATARAM
```
hzeller: The 'canoncial' way to do nested defines is to have the # start at the beginning, then some indentation. So something like
```
#if PRU_NUM == 0
#  define PRU_DATARAM ...  // notice the space between '#' and define.
#  define ....
#else
#  define PRU_DATARAM ...
#  define ...
#endif
```

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+
+bool UioPrussInterface::Init() {
+  tpruss_intc_initdata pruss_intc_initdata = PRUSS_INTC_INITDATA;
+  prussdrv_init();
+
+  /* Get the interrupt initialized */
+  int ret = prussdrv_open(PRU_EVTOUT_0);  // allow access.
+  if (ret) {
+    Log_error("prussdrv_open() failed (%d) %s\n", ret, strerror(errno));
+    return ret;
```
hzeller: Make sure to convert that to a proper boolean. The prusdrv_open() returns something that returns '0' on success and some error code otherwise.

Here, you want to indicate error by returning 'false'.

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+  }
+  prussdrv_pruintc_init(&pruss_intc_initdata);
+  return 0;
```
hzeller: Here, you want to return 'true'

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+}
+
+bool UioPrussInterface::AllocateSharedMem(void **pru_mmap, const size_t size) {
+  prussdrv_map_prumem(PRU_DATARAM, pru_mmap);
+  if (*pru_mmap == NULL) {
+    Log_error("Couldn't map PRU memory.\n");
+    return 1;
```
hzeller: Same here. This should return 'false'.

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+  }
+  bzero(*pru_mmap, size);
+  return 0;
```
hzeller: .. and this 'true'.

More below.

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+bool UioPrussInterface::StartExecution() {
+  prussdrv_pru_write_memory(PRU_INSTRUCTIONRAM, 0, PRUcode, sizeof(PRUcode));
+  prussdrv_pru_enable(PRU_NUM);
+  return 0;
+}
+
+unsigned UioPrussInterface::WaitEvent() {
+  prussdrv_pru_wait_event(PRU_EVTOUT0);
+  prussdrv_pru_clear_event(PRU_EVTOUT0, PRU_ARM_INTERRUPT);
+  return 0;
```
hzeller: Aren't we supposed to return number of events that have occured here ?

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
...
+}
+
+bool UioPrussInterface::Shutdown() {
+  prussdrv_pru_disable(PRU_NUM);
+  prussdrv_exit();
+  return 0;
```
hzeller: return true.

src/uio-pruss-interface.cc
```diff
@@ -0,0 +1,88 @@
+/* -*- mode: c++; c-basic-offset: 2; indent-tabs-mode: nil; -*-
+ * (c) 2013, 2014 Henner Zeller <h.zeller@acm.org>
```
hzeller: Since you started this file, you should add your name here also if you like.

hzeller:
Very nice, only some little confusion converting the old int-returning-C interface into a boolean-returning-C++ interface.

src/pru-motion-queue.cc
leonardoromor: (Solved)

src/pru-hardware-interface.h
leonardoromor: (Solved)

src/pru-hardware-interface.h
leonardoromor: (Solved): Placed it inside uio_pruss_interface

src/uio-pruss-interface.cc
leonardoromor: (Solved)

src/uio-pruss-interface.cc
leonardoromor: (Solved)

src/uio-pruss-interface.cc
leonardoromor: (Solved)

src/uio-pruss-interface.cc
leonardoromor: (Solved)

src/uio-pruss-interface.cc
leonardoromor: (Solved)

src/uio-pruss-interface.cc
leonardoromor: (Solved)

src/uio-pruss-interface.cc
leonardoromor: (Solved)

hzeller:
Looks good!

bigguiness:
All,

I may have just noticed a bug in:

unsigned UioPrussInterface::WaitEvent() {
  const unsigned num_events = prussdrv_pru_wait_event(PRU_EVTOUT0);
  prussdrv_pru_clear_event(PRU_EVTOUT0, PRU_ARM_INTERRUPT);
  return num_events;
}

prussdrv_pru_clear_event() is defined as:

int prussdrv_pru_clear_event(unsigned int host_interrupt, unsigned int sysevent);

WaitEvent() is passing the values backwards.

Maybe this is part of the race I am seeing.

Hartley

bigguiness:
Something broke with this commit.

I get this when I start machine-control:

machine-control: pru-motion-queue.cc:145: PRUMotionQueue::PRUMotionQueue(HardwareMapping*, PruHardwareInterface*): Assertion `init_result == 0' failed.

Rolling back to this commit works:

commit 6b5057b4c5627705e4266f251e5b4945f10677e0
Author: H Hartley Sweeten <hsweeten@visionengravers.com>
Date:   Tue Jul 26 14:08:37 2016 -0700

    Improve planner for threshold angle acceleration.

I just started looking into the failure.

Hartley

hzeller:
So the finding with the `prussdrv_pru_clear_event()` is interesting @bigguiness : the documentation indeed says first the interrupt, then the event.

But then all of their example code does the reverse:
https://github.com/beagleboard/am335x_pru_package/blob/master/pru_sw/example_apps/PRU_memAccessPRUDataRam/PRU_memAccessPRUDataRam.c#L148

Need to look at the implementation what the `prussdrv_pru_clear_event()` _actually_ is doing...

hzeller:
So there were two problems in this patch
1) there was some reverse logic going on essentially making the system immediately assert() fail. Fixed in https://github.com/hzeller/beagleg/commit/8ecd1621579a08cf98225847d07febb0a35c0f3c

2) The waiting for event was hanging, because it internally was reading from a wrong file-descriptor.
 instead of PRU_EVTOUT_0, the constant PRU_EVTOUT0 was used (which apparently is a thing internally in the PRU headers, with a different value). Fixed in https://github.com/hzeller/beagleg/commit/57d5ac13bb25ef5183ce241053fa5cd9624970dd

However, the PRU_EVTOUT_0 was used previously, so other hangs we might've seen must be some other reason.

PR #17, opened 2016-08-04
something is broken, but the clamping test now works even with differ…

leonardoromor:


hzeller:
I have now added the four test-cases like Hartley  described in his mail which might help more experiments.
I briefly looked if I can integrate the gcode-machine-control_test in your pull request #10 , but looks like this requires `clamp_to_limits()` first.

leonardoromor:
Hello, 
now all the tests are working, moreover, changing the ratios between the axes also do make the test pass. A similar clamping should be applied in the same way, for the accelerations. I will prepare another pull request if you wish , but it's just to apply the clamp_to_limits() function in the same way as for the speeds.

leonardoromor:
I also updated the gitignore to ignore planner_test.

hzeller:
Merged in 07c3bdca07f1254c32200c89b0bb2bccea470444

PR #19, opened 2016-09-02
added test showing which kind of configuration breaks the target spee…

lromor:
…d > 0 assert.

Probably this test can be extend with also other axes combinations.

### DIFF
--- a/src/gcode-machine-control_test.cc
+++ b/src/gcode-machine-control_test.cc
@@ -83,6 +83,14 @@ private:
 
   int errors_;
 };
+
+class DummyMotorOps : public MotorOperations {
+public:
+  DummyMotorOps() {}
+  virtual void Enqueue(const LinearSegmentSteps &param) {}
+  virtual void MotorEnable(bool on) {}
+  virtual void WaitQueueEmpty(){}
+};
 }
 
 class Harness {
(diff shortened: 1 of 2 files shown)


lromor:
I guess, the main reason here is this: https://github.com/hzeller/beagleg/blob/3b6c66b82d5ff89dfba6b30e70bc3aa2e8dc3f3a/src/gcode-machine-control.cc#L198

I would have prepared a patch, but since the I don't know the standard behavior or what do you expect as default feedrate (it has to take which value? a default one in mm/s? the slowest axis? the maximum euclidean speed? what if all the euclidean axes are 0 and only the A axis is not? Is it enough to put a "if the max_feedrate is 0, pick up the next"?

hzeller:
Well, some reasonable default that does not make the machine explode. In the case of Euclid, it could take all axes into account and do 10%. If it is still zero afterwards, then make it 1mm/s or something non-zero.

Note, the place you are pointing out is just for the case when someone starts using the machine and forgot to give a feedrate. The machine should actually complain probably instead of jugging along with 10% of the speed of some arbitrary axis.

src/gcode-machine-control_test.cc
```diff
@@ -112,6 +120,33 @@ class Harness {
...
+  HardwareMapping hardware;
+  DummyMotorOps motor_ops;
+
+  GCodeMachineControl *machine_control;
+  machine_control = GCodeMachineControl::Create(config, &motor_ops,
+                                                &hardware,
+                                                NULL,   // spindle
+                                                NULL);  // msg-stream
+  assert(machine_control != NULL);
+  // Move to pos 100, then 200, use default feedrate.
```
hzeller: There is no real default feedrate for GCode. The machine should actually report an error if you haven't given a feedrate on your first move.

The fact that BeagleG has one, is just because early testing was easier. We should report an error instead.
(or, if we want to be fancy, could have a default feedrate in the configuration, but not sure if this is not similarly obscuring the gcode-file error as it is now).

hzeller:
So we should rather report a problem if some gcode file relies on a default feedrate. On the first move, someone _has_ to give a feedrate.

lromor:
Perfect :) So I will log an error and wait for a feedrate to be set in order to start any movement.

lromor:
Ok, tell me if it makes sense.
-l

hzeller:
Submitted in b4641be0caf80a56207b7abd6334a04cbc99daad

PR #22, opened 2016-09-21
Update absolute position

lromor:


hzeller:
Thanks, that looks pretty good already. I have some concerns regarding unnecessary computation in the critical phases when we enqueue things in the history list, so made some suggestions there.

Essentially: if we always record in the history what at the _end_ of that the positions should look like, we can make the whole `UpdateAbsolutePosition()` implementation a lot cheaper, and we only ever would need to call it when we call `GetMotorLoops()` (which should be called GetMotorPosition()), not at enqueue time.

lromor:
wow. Nice solution

lromor:
Now the code should be simpler, shorter and faster.  Thanks,

lromor:
if possible, could you add to .gitignore also simple-lexer_test?

hzeller:
Looks much nicer. There are still a couple of smallish comments left, but I think we're mostly ready to merge.

hzeller:
I now added all *_test in .gitignore

src/motion-queue.h
```diff
@@ -73,6 +73,14 @@ struct MotionSegment {
 #endif
 } __attribute__((packed));
 
+// Layout of the status register
+// Assuming atomicity of 32 bit boundaries
+// This 32 bit value will be a copy of the R28 register of the PRU
```
lromor: Solved()

src/motion-queue.h
```diff
@@ -117,6 +132,15 @@ class PRUMotionQueue : public MotionQueue {
 
   volatile struct PRUCommunication *pru_data_;
   unsigned int queue_pos_;
+
+  // Shadow Queue
+  void RegisterHistorySegment(MotionSegment *element);
+  void UpdateAbsolutePosition();
+
+  struct HistorySegment *shadow_queue_;
+  int32_t absolute_pos_loops_[MOTION_MOTOR_COUNT];
```
lromor: (Solved)

src/motion-queue.h
```diff
@@ -117,6 +132,15 @@ class PRUMotionQueue : public MotionQueue {
...
+  int32_t segment_partials_[MOTION_MOTOR_COUNT];
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,80 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
 }
 #endif
 
+struct HistorySegment {
+  uint32_t total_loops;
+  uint32_t fractions[MOTION_MOTOR_COUNT];
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,80 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+  uint8_t direction_bits;
+};
+
+void PRUMotionQueue::RegisterHistorySegment(MotionSegment *element) {
+  const unsigned int last_insert_position = (queue_pos_ - 1) % QUEUE_LEN;
+  struct HistorySegment *new_slot = &shadow_queue_[last_insert_position];
+
+  bzero(new_slot, sizeof(struct HistorySegment));
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,52 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+struct HistorySegment {
+  uint32_t fractions[MOTION_MOTOR_COUNT];
+  uint32_t cumulative_loops[MOTION_MOTOR_COUNT];
+  uint8_t direction_bits;
+};
+
+void PRUMotionQueue::RegisterHistorySegment(MotionSegment *element) {
+  const uint64_t max_fraction = 0xFFFFFFFF;
+  const unsigned int last_insert_index = (queue_pos_ - 1) % QUEUE_LEN;
+  const HistorySegment previous
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,52 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+    = shadow_queue_[(last_insert_index - 1) % QUEUE_LEN];
+
+  const uint32_t total_loops
+    = element->loops_accel + element->loops_travel + element->loops_decel;
+  const uint8_t direction_bits = element->direction_bits;
+
+  HistorySegment *new_slot = &shadow_queue_[last_insert_index];
+
+  for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
```
lromor: (Solved) Ok

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,80 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+
+  new_slot->direction_bits = element->direction_bits;
+
+  for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
+    new_slot->fractions[i] = element->fractions[i];
+  }
+}
+
+static void fractions_to_loops(const struct HistorySegment &segment,
+                               int32_t *motors_loops) {
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,52 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+                                    sign * loops;
+  }
+
+  new_slot->direction_bits = element->direction_bits;
+}
+
+void PRUMotionQueue::GetMotorsLoops(int32_t *absolute_pos_loops) {
+  const uint64_t max_fraction = 0xFFFFFFFF;
+  const struct QueueStatus status = *(struct QueueStatus*) &pru_data_->status;
+  const HistorySegment current = shadow_queue_[status.index];
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,80 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+    const int sign = (segment.direction_bits >> i) & 1 ? -1 : 1;
+    uint64_t loops = (uint64_t) segment.fractions[i] * (uint64_t) segment.total_loops;
+    loops /= max_fraction;
+    motors_loops[i] = sign * loops;
+  }
+}
+
+void PRUMotionQueue::UpdateAbsolutePosition() {
+  const struct QueueStatus status = *(struct QueueStatus*) &pru_data_->status;
+  struct HistorySegment current = shadow_queue_[status.index];
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,80 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+  const int shadow_queue_delay
+    = (status.index + QUEUE_LEN - back_shadow_queue_) % QUEUE_LEN;
+  int32_t motors_loops[MOTION_MOTOR_COUNT];
+
+  for (int i = 0; i < shadow_queue_delay; ++i) {
```
lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -107,6 +183,12 @@ void PRUMotionQueue::Enqueue(MotionSegment *element) {
 
   // Fully initialized. Tell busy-waiting PRU by flipping the state.
   queue_element->state = state_to_send;
+
+  // Update the HistoryQueue before registering a new slot.
+  UpdateAbsolutePosition();
```
lromor: (Solved)

src/pru-motion-queue.cc
hzeller: Maybe add a comment that we don't forget that this still needs attention.

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,53 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
 #endif
 
+// Store the required informations needed to backtrack the absolute position.
+struct HistorySegment {
+  uint32_t fractions[MOTION_MOTOR_COUNT];
+  uint32_t cumulative_loops[MOTION_MOTOR_COUNT];
+  uint8_t direction_bits;
+};
+
+void PRUMotionQueue::RegisterHistorySegment(const MotionSegment *element) {
```
hzeller: If we're not modifying the MotionSegment, maybe we should pass it via const reference.

My rules I use for pointer vs. reference:
  Pointer: when we want to modify something.
  const Pointer: when we don't want to modify things, but if the parameter can be nullptr; so in that case we need to distinguish if if it is there or not.
  const Reference: when we don't want to modify things and if the thing is always there (so cannot be null).

(regular reference as a parameter without const: almost never: it has too surprising effects.)

lromor: (Solved)

src/pru-motion-queue.cc
```diff
@@ -79,6 +81,52 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
+    = element->loops_accel + element->loops_travel + element->loops_decel;
+  const uint8_t direction_bits = element->direction_bits;
+
+  HistorySegment *new_slot = &shadow_queue_[last_insert_index];
+
+  for (int i = 0; i < MOTION_MOTOR_COUNT; ++i) {
+    const uint32_t fraction = element->fractions[i];
+    new_slot->fractions[i] = fraction;
+    const int sign = (direction_bits >> i) & 1 ? -1 : 1;
+    uint64_t loops = (uint64_t) fraction * (uint64_t) total_loops;
```
lromor: (Solved)

src/motor-interface-pru.p
```diff
@@ -202,14 +216,30 @@ QUEUE_READ:
 	ZERO &mstate, SIZE(mstate)	; clear the motor states
 	ZERO &r3, 4			; initialize delay calculation state register.
 
+	;; STATUS REGISTER
+	;; ! We are assuming that writing the 4 bytes status register is atomic
+	;; and we guarantee that the bottom three bytes are all zero so we just need
+	;; to sum up the 3 loop counters. The upper bound of this sum will always be
+	;; less than 2^18. At each loop executed this counter is decreased of one unit.
```
lromor: (Solved)

src/motor-interface-pru.p
```diff
@@ -28,6 +28,7 @@
 #define CONST_PRUDRAM	   C24
 
 #define QUEUE_ELEMENT_SIZE (SIZE(QueueHeader) + SIZE(TravelParameters))
+#define STATUS_VALUE (QUEUE_LEN * QUEUE_ELEMENT_SIZE)
```
lromor: (Solved)

src/motion-queue.h
```diff
@@ -91,6 +99,10 @@ class MotionQueue {
 
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
+
+  // Fill the argument with the current absolute position in loops
+  // for each motor.
+  virtual void GetMotorsLoops(int32_t *absolute_pos_loops) = 0;
```
lromor: (Solved)

lromor: (Solved)

src/motor-interface-pru.p
lromor: (Solved)

src/motor-interface-pru.p
```diff
@@ -167,13 +168,26 @@ calc_decel:
 DONE_CALCULATE_DELAY:
 .endm
 
+;;; This macro decrease the counter that holds the overall number of loops left
+;;; to be performed and then it push it in the PRU DRAM status register.
```
lromor: (Solved)

src/pru-motion-queue.cc
lromor: (Solved)

hzeller:
So I tried this patch on my machine, but it starts hanging when it does a `G28` and reaches the endstop. I have not dug into details yet why this is happening.
(this happens on the z-axis with home-pos = max in case that is relevant). Can you have a look ?

BTW, I changed little things so that it passes my test with various compilers and valgrind. Attached a patch relative to your change (and I had to name it *.txt instead of *.patch as github then complains that it can't deal with that file type. D'oh).

[valgrind-proof.txt](https://github.com/hzeller/beagleg/files/505631/valgrind-proof.txt)

lromor:
it seems the subtract thing is making problems, from what I saw with the debugger there's a no wait loop, causing him do underflow.. or something like that

lromor:
So, for X and Y works right?

hzeller:
I don't know, the Z axis is the first that was homed when it got stuck :)

lromor:
Ok, I think now should be fixed :)

hzeller:
Alright, seems to work. Pushed as submit https://github.com/hzeller/beagleg/commit/03950be17eec2490909d67048824e2cf50d7ed55
(with only minor changes to make it compile with various compilers and with uninitalized memory fixed as found by `make valgrind-test`).

Thanks!

PR #26, opened 2017-04-14
moving the shadow queue into motor-operations

lromor:


src/motor-operations.h
```diff
@@ -45,6 +46,17 @@ struct LinearSegmentSteps {
   int steps[BEAGLEG_NUM_MOTORS]; // Steps for axis. Negative for reverse.
 };
 
+// Struct used to return data about the currently executed steps
+// and the status of the auxes.
+struct RealtimeStatus {
```
hzeller: RealtimeStatus is not really a good name. It is something we get back when asking with `GetRealtimeStatus()`, but what we get back is fundamentally only really a current configuration of where the motors are.

We should probably have a struct not called RealtimeStatus, but PhysicalPosition or PhysicalStatus or something (naming not quite clear yet).
So say
```
struct PhysicalPosition {
    int steps[BEAGLEG_NUM_MOTORS];
    unsingned short aux_bits;
};
```
(the fact that there are aux bits makes 'position' a bit bad, maybe it should be called `PhysicalStatus`).

Then we can use this type in `LinearSegmentSteps`

```
struct LinearSegmentSteps {
    float v0;
    float v1;
    PhysicalPosition target;
}
```

.. and when asking the realtime position.
```
 virtual void GetRealtimeStatus(PhysicalPosition *status);
```

I think it helps in reducing the number of different types used, and it makes it more clear what the intend is.

src/motor-operations.h
```diff
@@ -62,22 +74,32 @@ class MotorOperations {  // Rename SegmentQueue ?
 
   // Wait, until all elements in the ring-buffer are consumed.
   virtual void WaitQueueEmpty() = 0;
+
+  // Get the absolute position and auxes status currently
+  // executed by the backend.
+  virtual void GetRealtimeStatus(RealtimeStatus *status) = 0;
 };
 
+struct HistorySegment;
```
hzeller: Is history segment something that should be in the toplevel namespace ? It looks like it is something that should be a private type member of MotionQueueMotorOperations.
You can just forward declare it after the `private:`.

src/pru-motion-queue.cc
```diff
@@ -216,5 +177,10 @@ bool PRUMotionQueue::Init() {
   }
   queue_pos_ = 0;
 
+  // The MSB 8 bits of the status register shoud represent the index
+  // of the last executed motion segment. If no motion segments has been
+  // yet pushed in the queue, the index should represent the index before
+  // the slow we would push, in this case -1.
```
hzeller: so, uhm, why not call it -1 then instead of 0xff ? If status.index is a char, it should set it to the same bit-pattern.

I'd probably not use this constant here, but have some
  constexpr int kInitializedIndexSentinel = 0xff;
or something. And then use this here and in the Status() function where you look at it.

src/pru-motion-queue.cc
```diff
@@ -82,54 +82,18 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
-    loops /= max_fraction;
-    (*absolute_pos_loops)[i] = current.cumulative_loops[i] - sign * loops;
+  if (executed_loops) *executed_loops = status.counter;
+  // Nothing has been pushed yet.
+  if (status.index == 0xff) {
+    *buffer_size = 0;
+    return;
   }
+  *buffer_size = (queue_pos_ - status.index) % QUEUE_LEN;
+  *buffer_size += !(*buffer_size) * QUEUE_LEN;
```
hzeller: what is this supposed to do ? You are derefencing buffer_size, interpret it as boolean, and if its value was zero (inverted 1), then you add QUEUE_LEN to buffer size, otherwise zero (a very complicated way of writing `*buffer_size ? 0 : QUEUE_LEN`).

This is not a very readable piece of expression.

src/pru-motion-queue.cc
```diff
@@ -82,54 +82,18 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
-    loops *= counter;
-    loops /= max_fraction;
-    (*absolute_pos_loops)[i] = current.cumulative_loops[i] - sign * loops;
+  if (executed_loops) *executed_loops = status.counter;
+  // Nothing has been pushed yet.
+  if (status.index == 0xff) {
+    *buffer_size = 0;
+    return;
   }
+  *buffer_size = (queue_pos_ - status.index) % QUEUE_LEN;
```
hzeller: What if queue pos is less than status.index (isn't it wrapping around) ? Then you modulo a negative number which usually is not resulting in what you want. So you want to add QUEUE_LEN first to queue_pos_ to have that covered.

src/pru-motion-queue.cc
```diff
@@ -82,54 +82,18 @@ static void DumpMotionSegment(volatile const struct MotionSegment *e,
...
-                                    + sign * loops;
-  }
-
-  new_slot->direction_bits = element.direction_bits;
-}
-
-void PRUMotionQueue::GetMotorsLoops(MotorsRegister *absolute_pos_loops) {
-  const uint64_t max_fraction = 0xFFFFFFFF;
+void PRUMotionQueue::Status(uint32_t *executed_loops,
+                            unsigned int *buffer_size) {
```
hzeller: So looking at the implemtation below it does look like you want to show the difference in how much things are ahead of us, but then if nothing is ahead of us, you return QUEUE_LEN ? This is confusing at best.
What am I missing ?

src/motor-operations.cc
```diff
@@ -85,11 +86,39 @@ static char test_acceleration_ok(float acceleration) {
 }
 #endif
 
+typedef struct HistoryPositionInfo {
+  HistoryPositionInfo () : position_steps(0), sign(1) {}
+  int position_steps;
+  uint32_t fraction;
+  short sign;
```
lromor: solved

src/motor-operations.cc
```diff
@@ -152,6 +187,39 @@ void MotionQueueMotorOperations::EnqueueInternal(const LinearSegmentSteps &param
...
+void MotionQueueMotorOperations::GetRealtimeStatus(RealtimeStatus *status) {
+  // Shrink the queue
+  unsigned int buffer_size;
+  uint32_t loops;
+  backend_->Status(&loops, &buffer_size);
+  assert(buffer_size <= shadow_queue_->size());
+  if (shadow_queue_->size() > 1) shadow_queue_->resize(buffer_size);
+
+  // Get the last element
+  const HistorySegment hs = shadow_queue_->back();
```
lromor: solved

src/motion-queue.h
```diff
@@ -105,9 +105,10 @@ class MotionQueue {
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
 
-  // Fill the argument with the current absolute position in loops
-  // for each motor.
-  virtual void GetMotorsLoops(MotorsRegister *absolute_pos_loops) = 0;
+  // Return the number of motion segments enqueued after
+  // the one currently executing and its already executed loops.
+  // buffer_size will never be 0 except if nothing has been yet enqueued.
+  virtual void Status(uint32_t *executed_loops, unsigned int *buffer_size) = 0;
```
lromor: > The description is not entirely clear.
> You describe that 'returns the number of motion segments', but then the name of the variables does not reflect that. As reader, I could infer that it is buffer_size, but it is not entirely clear (later down in the code it become clear, that this is an ever increasing value until the queue has reached some equilibrum. I think this needs to be better explained in the method description or solved otherwise.
> (was this the question about sharing information about the queue size ?)

Yes precisely, 
the layer above motion-queue should be informed about the number of segments it has to keep hence not yet executed. 

So the when the queue beneath is full  (ie is executing the slot with index 5 and the last enqueued segment is the 4th) it returns 16, because I need to rollback of 16 segments. 

When instead all the segments are executed, motor-operations should keep just one segment so buffer_size is 1, and the number of loops returned will be 0 because the pru is waiting for a new slot to be enqueued.

src/motor-operations.h
lromor: Solved()

src/pru-motion-queue.cc
lromor: that was confusing me a lot as well, I don't know if it's for the same reason.
So, if queue_pos_ == execution_index_  is something possible **only** if the queue is full, not empty. This is because queue_pos_ represent the next slot to be filled:

`&pru_data_->ring_buffer[queue_pos_++];`

src/pru-motion-queue.cc
lromor: somehow I thought that if you can avoid an explicit conditional statement you can slightly increase performance of the statement. 

But this is not backed-up by strong facts.

src/pru-motion-queue.cc
hzeller: I think we should find a solution that does not require this confusing passing of the buffer size. Even if we have to give the upstream user the buffer size in the constructor (and hence have to use QUEUE_LEN outside of the 'hardware zone'), this would be better.

src/pru-motion-queue.cc
lromor: Solved()

src/pru-motion-queue.cc
hzeller: Yes, sometimes such tricks can result in not having a branch. But
  * Never do premature optimization. Only ever consider manual optimization in code that has shown to be too slow in a critical path.
  * If you do things like this: _measure its impact_. Some (most?) well meaning manual optimizations can be worse (e.g. here you replace a MOV with a multiplication which typically is more expensive, so a positive impact has to be proven - and also is platform dependent).
  * If you do "clever" things that are not necessarily obvious for the reader, you have to comment and explain them, including reasons for need of optimization and measurement results of the impact.
  * In general: let the compiler be "clever". If it sees a ?: operation, it can do much better optimization (e.g. if it knows that multiplication is cheap, it can do what you did automatically). If you obfuscate your intend, it is harder for the compiler to do its job.

src/pru-motion-queue.cc
lromor: Solved()

src/pru-motion-queue.cc
lromor: I guess it would be better to pass QUEUE_LEN in the constructor especially for test pourposes (ie tests for different queue sizes). It could be done in a patch after this, what do you prefer?

hzeller: Given that it probably would be more work to come up with a good explanation about the meaning of the buffer_size that we can put in the comment, it would be better if we do this change now.

src/motor-operations.h
lromor: Solved()

src/motor-operations.cc
```diff
@@ -85,11 +86,39 @@ static char test_acceleration_ok(float acceleration) {
...
+
+// Store the required informations needed to backtrack the absolute position.
+struct HistorySegment {
+  HistoryPositionInfo pos_info[MOTION_MOTOR_COUNT];
+  unsigned short aux_bits;
+};
+
+MotionQueueMotorOperations::MotionQueueMotorOperations(MotionQueue *backend)
+                                                       : backend_(backend) {
+  shadow_queue_ = new std::deque<struct HistorySegment>();
```
lromor: Solved()

src/pru-motion-queue.cc
lromor: Solved()

src/motion-queue.h
```diff
@@ -105,9 +105,11 @@ class MotionQueue {
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
 
-  // Fill the argument with the current absolute position in loops
-  // for each motor.
-  virtual void GetMotorsLoops(MotorsRegister *absolute_pos_loops) = 0;
+  // Return the number of motion segments enqueued after
+  // the one currently executing and its already executed loops.
+  // buffer_size will never be 0 except if nothing has been yet enqueued.
+  virtual void GetExecutionProgress(uint32_t *loop_progress_working_segment,
```
hzeller: In the offline discussion, we had already had some better concept. I think something like
```
   virtual int GetPendingElements(uint32_t *head_item_progress);
```
or something (naming needs improvement). Essentially, since this is expected to behave as a queue, we should have functions that reflect that. This naturally would reflect how much there is to do; if the number returned is non-zero, the queue is working on something and the head_item_progress would tell us how much.

src/motion-queue.h
```diff
@@ -105,9 +105,10 @@ class MotionQueue {
   // Shutdown. If !flush_queue: immediate, even if motors are still moving.
   virtual void Shutdown(bool flush_queue) = 0;
 
-  // Fill the argument with the current absolute position in loops
-  // for each motor.
-  virtual void GetMotorsLoops(MotorsRegister *absolute_pos_loops) = 0;
+  // Return the number of motion segments enqueued after
+  // the one currently executing and its already executed loops.
+  // buffer_size will never be 0 except if nothing has been yet enqueued.
```
hzeller: This comment is now not accurate anymore (there is no buffer_size).

Also, it is not quite clear from the description: While something is executing, what does this method return ? I would expect '1' (as there is still one element being worked on) and the head item progress is set to some value. While Zero would be returned when the queue is fully executed and empty (and in that case, we don't expect head_item_progress to be updated for instance).

src/motor-operations.cc
```diff
@@ -85,11 +86,40 @@ static char test_acceleration_ok(float acceleration) {
...
+  unsigned short aux_bits;
+};
+
+MotionQueueMotorOperations::
+  MotionQueueMotorOperations(MotionQueue *backend)
+  : backend_(backend),
+    shadow_queue_(new std::deque<struct HistorySegment>()) {
+  // Initialize the history queue.
+  struct HistorySegment new_hs = {};
+  shadow_queue_->push_front(new_hs);
```
hzeller: Wouldn't this work by directly do shadow_queue_->push_front({});
?

Another thing that I am wondering: here we are essentially resetting the target position to be {0,0,...}.
If the machine is not homed yet in the beginning so needs to go backwards - then after homing the absolute positions in our history queue are negative and such, but this is working fine ?

src/motor-operations.cc
```diff
@@ -85,11 +86,40 @@ static char test_acceleration_ok(float acceleration) {
 }
 #endif
 
+typedef struct HistoryPositionInfo {
+  HistoryPositionInfo () : position_steps(0), sign(1) {}
+  int position_steps;
```
hzeller: Can this be more described ?
I think the 'position_steps' is the absolute machine position at the end of the move (important to describe as this is different from the delta steps the queue gets).
So the fraction is also good to describe (1/0xffffffff-th of the difference between the previous step and now)

src/motor-operations.cc
```diff
@@ -85,11 +86,40 @@ static char test_acceleration_ok(float acceleration) {
...
+MotionQueueMotorOperations::~MotionQueueMotorOperations() {
+  delete shadow_queue_;
+}
+
 void MotionQueueMotorOperations::EnqueueInternal(const LinearSegmentSteps &param,
                                                  int defining_axis_steps) {
   struct MotionSegment new_element = {};
   new_element.direction_bits = 0;
 
+  // Get the last history segment pushed
```
hzeller: In comments: Don't describe what the code is already saying ('get last history segment'), but rather why that is. Here, because the final position is relative to the previous position. So something like
   // The new segment is based on the previous position
or similar.

src/motor-operations.cc
```diff
@@ -152,6 +188,38 @@ void MotionQueueMotorOperations::EnqueueInternal(const LinearSegmentSteps &param
...
+  shadow_queue_->resize(buffer_size);
+}
+
+void MotionQueueMotorOperations::GetRealtimeStatus(PhysicalStatus *status) {
+  // Shrink the queue
+
+  uint32_t loops;
+  const unsigned int buffer_size = backend_->GetPendingElements(&loops);
+  assert(buffer_size <= shadow_queue_->size());
+  if (shadow_queue_->size() > 1) shadow_queue_->resize(buffer_size);
```
hzeller: So this resize() business seems a bit hand-wavy to me (here, and in the EnqueInternal()).

This is because you want to try hard to avoid knowing the queue-size of the underlying MotionQueue so do all kinds of observational second guessing here. But it would be so much easier (and easier for someone reading the code) if you just kept track of it with a ring-buffer.  So I would suggest that either
  * you pass the size of the queue len to the constructor of the MotionQueueMotorOperations
  * or have a size_t MotionQueue::GetFixedInternalQueueSize() const; that operations downstream (such as this MotionQueueMotorOperations) can query and set-up their fixed

src/motor-operations.h
```diff
@@ -45,6 +46,17 @@ struct LinearSegmentSteps {
...
 };
 
+// Struct used to return data about the currently executed steps
+// and the status of the auxes.
+struct PhysicalStatus {
+  // Absolute position in steps
+  int pos_steps[BEAGLEG_NUM_MOTORS];
+  // Auxes status
+  unsigned short aux_bits;
+  // A Tag? It may be useful to add a tag to map back
```
hzeller: Yes, that sounds like a good idea in the future. We would need to provide that tag when we Enqueue() the LinearSegmentStep.
(something for future work).

src/motor-operations.h
```diff
@@ -45,6 +46,17 @@ struct LinearSegmentSteps {
   int steps[BEAGLEG_NUM_MOTORS]; // Steps for axis. Negative for reverse.
 };
 
+// Struct used to return data about the currently executed steps
+// and the status of the auxes.
+struct PhysicalStatus {
+  // Absolute position in steps
+  int pos_steps[BEAGLEG_NUM_MOTORS];
```
hzeller: Like above, it might be interesting to clarify what 'absolute' means in this case. When we start with an un-homed machine, then home, the absolute position might be some value relative to where the machine started ?

src/motor-operations.h
```diff
@@ -62,22 +74,31 @@ class MotorOperations {  // Rename SegmentQueue ?
 
   // Wait, until all elements in the ring-buffer are consumed.
   virtual void WaitQueueEmpty() = 0;
+
+  // Get the absolute position and auxes status currently
+  // executed by the backend.
+  virtual void GetRealtimeStatus(PhysicalStatus *status) = 0;
```
hzeller: We have several implementations (in tests or other simulations), that can't provide the PhysicalStatus and dont' update the value. Maybe we should have this method return a boolean indicating if the value was updated.

src/motion-queue.h
lromor: (Solved)

src/motion-queue.h
```diff
@@ -105,9 +105,10 @@ class MotionQueue {
...
+  virtual unsigned int GetPendingElements(uint32_t *head_item_progress) = 0;
```
lromor: (Solved)

src/motor-operations.cc
lromor: (Solved)

src/motor-operations.cc
lromor: > If the machine is not homed yet in the beginning so needs to go backwards - then after homing the absolute positions in our history queue are negative and such, but this is working fine ?

I preferred to keep the position returned absolute also in time so that motor operations do not actually know anything about homing. They just return what they know, the position of the motors since beagleg started.

What I did was retranslating the position using some homing_offset[] array.

src/motor-operations.cc
lromor: (Solved)

src/motor-operations.h
lromor: (Solved)

src/motor-operations.h
lromor: (Solved=Removed the comment)

hzeller:
Ok, so let's see if we can get this review finally finished. Sorry for the delay, I was kindof distracted with some other project in-between ([LDGraphy](http://ldgraphy.org)).

We still have this one open question of where to test the queue, but I think it is fine if we submit this now and then evolve it over time. I don't want to stand in the way of this getting submitted.

There are still the one or other smaller remark though, please have a look if these still need addressing.

Since we have in the meantime several submits, can you provide a good description that we can put on the final merged submit for this change that describes what we do here ? (like describing the separation of the various queues).

Unfortunately, my CNC machine is still in re-wiring mode, so I ask the folks on the mailing list to test the current patch.

src/pru-motion-queue_test.cc
```diff
@@ -44,188 +44,122 @@ class MockPRUInterface : public PruHardwareInterface {
...
   MotorsRegister absolute_pos_loops;
-  MockPRUInterface *pru_interface = new MockPRUInterface();
-  HardwareMapping *hmap = new HardwareMapping();
-  PRUMotionQueue motion_backend(hmap, (PruHardwareInterface*) pru_interface);
+  MockPRUInterface pru_interface = MockPRUInterface();
+  HardwareMapping hmap = HardwareMapping();
+  PRUMotionQueue motion_backend(&hmap, (PruHardwareInterface*) &pru_interface);
 
-  motion_backend.Enqueue(&segment);
+  static struct MotionSegment segment = {0};
```
hzeller: mmh, I forgot, is there a reason why we have this static ?
(it looks like we had that before as well, but not sure why).

The single {0} will be warned if you compile with g++6:
CXX=g++-6 BEAGLEG_OPT_CFLAGS="-W -Wno-unused-parameter" make test

src/pru-motion-queue_test.cc
```diff
@@ -44,188 +44,122 @@ class MockPRUInterface : public PruHardwareInterface {
...
-
-  pru_interface->SimRun(1, 150);
-  motion_backend.GetMotorsLoops(&absolute_pos_loops);
-
-  EXPECT_THAT(expected, ::testing::ContainerEq(absolute_pos_loops));
-
-  delete pru_interface;
-  delete hmap;
+  pru_interface.SimRun(1, 0);
+  check_buffer_value(&motion_backend, 1);
```
hzeller: Q: So we are not testing anymore all the movements (like the MotorsRegister 'expected' thing). Are all these tests now in motor-operations_test ?
(looks like it, just double checking)

src/pru-motion-queue_test.cc
```diff
@@ -44,188 +44,122 @@ class MockPRUInterface : public PruHardwareInterface {
...
-  MotorsRegister absolute_pos_loops;
-  MockPRUInterface *pru_interface = new MockPRUInterface();
-  HardwareMapping *hmap = new HardwareMapping();
-  PRUMotionQueue motion_backend(hmap, (PruHardwareInterface*) pru_interface);
-
-  motion_backend.GetMotorsLoops(&absolute_pos_loops);
-
-  const MotorsRegister expected = {0, 0, 0, 0, 0, 0, 0, 0};
-  EXPECT_THAT(expected, ::testing::ContainerEq(absolute_pos_loops));
+static void check_buffer_value(PRUMotionQueue *motion_backend,
```
hzeller: Instead of this function call, I think it is better if wherever we call this function would directly call the EXPECT_EQ()

EXPECT_EQ(motion_backend.GetPendingElements(NULL), 0);

Because EXPECT_EQ() is a macro, it can log the right line number whenever this is happening (with the check_buffer_value() it would only complain about line number 69, which is less helpful)

src/motor-operations.h
lromor: (Solved)

src/pru-motion-queue_test.cc
lromor: (Solved)

src/pru-motion-queue_test.cc
lromor: (Solved)

I don't really remember either, but I  cannot think for a reason right now

src/pru-motion-queue_test.cc
lromor: exactly, now we evaluate that part in motor-operations.cc so it has been moved in motor-operations_test

PR #29, opened 2017-12-11
linking M114 gcode command to the pru motors status backend

lromor:
o connecting m114 to motor-operation's GetPhysicalStatus backend.
o fixing some comments.
o updating dummy and simulated GetPendingElements behavior.

### DIFF
--- a/src/hardware-mapping.h
+++ b/src/hardware-mapping.h
@@ -189,6 +189,11 @@ public:
   // motors in the LinearSegmentSteps
   void AssignMotorSteps(LogicAxis axis, int steps, LinearSegmentSteps *out);
 
+  // Given the logical axis and an array containing all the motors steps
+  // it returns true if the axis is mapped to a motor and axis_steps
+  // is filled with the corresponding number of steps.
+  bool AssignAxisSteps(LogicAxis axis, int *motors_steps, int *axis_steps);
+
   // -- Switch access
 
   // Returns which endstop trigger are available. Possible values are
--- a/src/planner.cc
+++ b/src/planner.cc
@@ -556,13 +556,16 @@ void Planner::Impl::bring_path_to_halt() {
 }
 
 void Planner::Impl::GetCurrentPosition(AxesRegister *pos) {
-  // TODO: this should use the MotorOperations::GetPhysicalStatus to get
-  // the very last postition where we are in.
-  assert(planning_buffer_.size() > 0);  // we always should have a current pos
-  const int *mpos = planning_buffer_[0]->position_steps;
+  PhysicalStatus physical_status;
+  motor_ops_->GetPhysicalStatus(&physical_status);
+  int axis_steps;
   for (const GCodeParserAxis a : AllAxes()) {
-    if (cfg_->steps_per_mm[a] != 0) {
-      (*pos)[a] = 1.0f * mpos[a] / cfg_->steps_per_mm[a];
+    const bool axis_mapped =
+      hardware_mapping_->AssignAxisSteps(a, physical_status.pos_steps, &axis_steps);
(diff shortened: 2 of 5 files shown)


src/planner.cc
```diff
@@ -556,13 +556,16 @@ void Planner::Impl::bring_path_to_halt() {
...
-  assert(planning_buffer_.size() > 0);  // we always should have a current pos
-  const int *mpos = planning_buffer_[0]->position_steps;
+  PhysicalStatus physical_status;
+  motor_ops_->GetPhysicalStatus(&physical_status);
+  int axis_steps;
   for (const GCodeParserAxis a : AllAxes()) {
-    if (cfg_->steps_per_mm[a] != 0) {
-      (*pos)[a] = 1.0f * mpos[a] / cfg_->steps_per_mm[a];
+    const bool axis_mapped =
+      hardware_mapping_->AssignAxisSteps(a, physical_status.pos_steps, &axis_steps);
```
hzeller: With the GetAxisSteps() this loop can be simplified to
```
 pos->zero();
  for (const GCodeParserAxis a : AllAxes()) {
      if (cfg_->steps_per_mm[a] != 0) {
              (*pos)[a] = GetAxisSteps(a, physical_status) / cfg_->steps_per_mm[a]);
      }
  }
```

src/hardware-mapping.h
```diff
@@ -189,6 +189,11 @@ class HardwareMapping {
   // motors in the LinearSegmentSteps
   void AssignMotorSteps(LogicAxis axis, int steps, LinearSegmentSteps *out);
 
+  // Given the logical axis and an array containing all the motors steps
+  // it returns true if the axis is mapped to a motor and axis_steps
+  // is filled with the corresponding number of steps.
+  bool AssignAxisSteps(LogicAxis axis, int *motors_steps, int *axis_steps);
```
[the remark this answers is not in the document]
lromor: (Solved)

src/planner.cc
lromor: (Solved)

hzeller:
So this results in weird results. Here I am moving to (100, 100, 100) and query `M114` in-between. It beautifully shows how the position is changing:

```
G1 X100 Y100 Z100 F100
M114
X:1.547 Y:1.548 Z:1.550 E:0.000 [ABS. MACHINE CUBE X:1.547 Y:1.548 Z:1.550] (Unsure: machine never homed!)
ok
M114
X:28.528 Y:28.531 Z:28.534 E:0.000 [ABS. MACHINE CUBE X:28.528 Y:28.531 Z:28.534] (Unsure: machine never homed!)
```

Now it arrived. I am asking again the position, expecting it to be (100, 100, 100) now:
```
M114
X:0.000 Y:0.000 Z:0.000 E:0.000 [ABS. MACHINE CUBE X:0.000 Y:0.000 Z:0.000] (Unsure: machine never homed!)
ok
```

Mmh, position **(0, 0, 0)** ? Let's try if the machine at least thinks it is at (100, 100, 100):

```
G1 X100 Y100 Z100 F100
ok
M114
X:0.000 Y:0.000 Z:0.000 E:0.000 [ABS. MACHINE CUBE X:0.000 Y:0.000 Z:0.000] (Unsure: machine never homed!)
ok
```
Machine does not move (as expected, because it is already there), but `M114` still shows (0, 0, 0).

Now let's move to 0 again
```
G1 X0
ok
M114
Connection closed by foreign host.
```

Oops, crashed. The last log output of machine-control at exit is:
```
machine-control: motor-operations.cc:205: virtual bool MotionQueueMotorOperations::GetPhysicalStatus(PhysicalStatus*): Assertion `buffer_size <= (int)shadow_queue_->size()' failed.
```

I am rolling back this change for now, so that we can look at the reason without rush.

PR #41, opened 2018-06-20
Detecting the anomaly acceleration in a segment

LeonBrov:
I found a problem with the wrong speed determining for the segment in the planner. This occurs with short segments of the movement,followed by a change in the direction of motion. This is typical for the case when a straight line goes into an arc or circle. In these segments, due to a high initial speed and a short braking path in the segment,there is an acceleration exceeding the permissible limits.  That as a result leads to an uncontrollable situation.

### DIFF
--- a/src/planner_test.cc
+++ b/src/planner_test.cc
@@ -403,6 +403,87 @@ TEST(PlannerTest, CornerMove_Shallow_NegativeAngle) {
   testShallowAngleAllStartingPoints(kThresholdAngle, kTestingAngle);
 }
 
+//This function compute acceleration based on the starting velocity, the ending velocity and the path in steps
+int acceleration_segment(float v0, float v1, int delta){
+  double acceleration;
+  acceleration=(v1*v1-v0*v0)/(2.0*abs(delta));
+  return (int)(acceleration);
+}
+// I found a problem with the wrong speed determining for the segment in the planner.
+// This occurs with short segments of the movement,followed by a change in the direction of motion.
+// This is typical for the case when a straight line goes into an arc or circle. In these segments,
+// due to a high initial speed and a short braking path in the segment,there is an acceleration exceeding the permissible limits.
+// That as a result leads to an uncontrollable situation.
+TEST(PlannerTest, DetectOutRangeAcceleration) {
+  MachineControlConfig *config = new MachineControlConfig();
+  InitTestConfig(config);
+  config->steps_per_mm[AXIS_X] = 1000;
+  config->steps_per_mm[AXIS_Y] = 1000;
+  config->acceleration[AXIS_X] = 100;  // mm/s^2
+  config->acceleration[AXIS_Y] = 100;  // mm/s^2
+  PlannerHarness plantest(2.5,0, config);
+  float speed=10000;
+  AxesRegister pos;
+  pos[AXIS_X] = 60;
+  pos[AXIS_Y] = 250;
+  plantest.Enqueue(pos, speed); //long movement
+  pos[AXIS_X] = 60;
+  pos[AXIS_Y] = 100;
+  plantest.Enqueue(pos, speed); //long movement
+  pos[AXIS_X] = 60;
+  pos[AXIS_Y] = 99.9;
+  plantest.Enqueue(pos, speed); //short movement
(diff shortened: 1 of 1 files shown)


hzeller:
Thanks for providing example code that exposes the problem! Having test
code is always a good way to verify later if things have been fixed.

We know that at this point, the planner is not perfect yet and is a little
too simplistic at times, and indeed for some short segments we have some
optimizations that can violate the constraints.
(Adding Hartley as CC, as he is thinking about the tradeoffs between
'optimal movement' vs. 'reasonable fast machine' tradeoffs).

Good news is, that I'll allocated some time this summer to bring the
planner into a better shape.

If you would like to help out in the process, subscribe to beagleg-dev
where we sometimes discuss the details.

Thanks,
  Henner.

src/planner_test.cc
```diff
@@ -403,6 +403,87 @@ TEST(PlannerTest, CornerMove_Shallow_NegativeAngle) {
...
+  pos[AXIS_X] = 60;
+  pos[AXIS_Y] = 99.8;
+  plantest.Enqueue(pos, speed);//short movement
+  pos[AXIS_X] = 60;
+  pos[AXIS_Y] = 99.7;
+  plantest.Enqueue(pos, speed);  //short movement
+  pos[AXIS_X] = 60.1;
+  pos[AXIS_Y] = 99.6;
+  plantest.Enqueue(pos, speed);  //change direction angle 45 degree, this is the problematic part of moving
+  pos[AXIS_X] = 60.2;
```
bigguiness: Finally got a chance to look at this...

The current planner only looks at three "positions" to determine the acceleration profile:
1) the current position
2) the position we want to move to
3) the next upcoming position

The 1st Enqueue would see these positions:
1) X0 Y0    2) X60 Y250    3) X60 Y100
This will result in an accel segment to speed, a travel segment at speed, and a decel segment to 0 because the Y axis is turning around.

The next Enqueue gets these positions:
1) X60 Y250    2) X60 Y100    3) X60 Y99.9
These are codirectional so there will be an accel segment to speed and a travel segment at speed.

The next Enqueue:
1) X60 Y100    2) X60 Y99.9    3) X60 Y99.8
Again these are codirectional so we should just have a travel segment at speed.

Next:
1) X60 Y99.9    2) X60 Y99.8    3) X60 Y99.7
Again these are codirectional so we should just have a travel segment at speed.

Next:
1) X60 Y99.8    2) X60 Y99.7    3) X60.1 Y99.6
The next upcoming position is a 45 degree direction change. The planner will detect this and try to decelerate the motion to 0. The problem is we are already running at speed. With the 100 mm/s^2 acceleration there is no way the speed can be decelerated to 0 in the 0.1 distance. Your debug output actually shows this:

segment[5] v0=173320.515625  v1=0.000000    lenghtX=0   lenghtY=-100   accelX=0  accelY=-150200002 
planner_test.cc:502: Failure
Expected: (config->acceleration[AXIS_Y] * config->steps_per_mm[AXIS_Y] * 1.003) >= (abs(ay)), actual: 100300 vs 150200002

According to that you would need an accel of 150200002 in order to decel from speed to 0 in the given distance.

I don't want to look it up right now ;-), but Henner probably knows the math to determine the exact distance needed to decel from a 10000 speed to a 0 speed with the accel rate in your test code.

I was able to get the test to pass by replacing the X60 Y100, X60 Y99.9, and X60 Y99.8 moves with an X60 Y180 move. That makes the move before the 45 degree turn look like this:
1) X60 Y180    2) X60 Y99.7    3) X60.1 Y99.6
The planner still sees the upcoming 45 degree turn. We are still running at full speed but now we have enough distance to decel to 0 to make the turn.

I think the only real "fix" for this is to increase the look ahead in the planner to that it can detect earlier that a decel is going to be needed.

Hartley

hzeller:
So, uhm, a little late, but finally I have merged this PR as a planner fix is right around the corner (did ran `.github/bin/run-clang-format.sh` on it, as this had been added recently)

I have `#if 0` out the test for now (as it is still failing), but Lenoardo is actually fixing the planner currently (#44) so when that is merged, this test can be enabled.

hzeller:
Merged manually (with the `#if 0` for now), so this can be closed.

PR #44, opened 2021-05-03
Added simple test that ensures accelerations constraints are met.

lromor:


src/planner_test.cc
```diff
@@ -170,6 +173,40 @@ class PlannerHarness {
...
+    sqrtf(axes[0] * axes[0] + axes[1] * axes[1] + axes[2] * axes[2]);
+
+  for (int i = 0; i <= AXIS_Z; ++i) {
+    const GCodeParserAxis axis = (GCodeParserAxis)i;
+    if (segment.steps[axis]) {
+      const float accel =
+        fabs_accel(segment.v0, segment.v1, segment.steps[axis]) * axes[i] /
+        (config.steps_per_mm[axis] * hypotenuse);
+      const float max_accel = config.acceleration[axis];
+      EXPECT_TRUE(fabs(accel - max_accel) < epsilon)
```
hzeller, line 202: there is also an `EXPECT_NEAR()` macro in google test.

lromor: (solved using EXPECT_NEAR_REL)

hzeller:
Looks like there is a conflict in one of the files; the last change in the file was in #47 - maybe things got messed up in working in two branches ?

src/planner.cc
```diff
@@ -18,16 +18,26 @@
  */
 #include "planner.h"
 
+#include <bits/stdint-uintn.h>
```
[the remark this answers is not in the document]
lromor: Thanks! Yeah sometimes emacs with lsp-mode automatically imports the wrong headers when he doesn't find some type.
(solved)

src/planner.cc
```diff
@@ -38,17 +47,77 @@ namespace {
...
+    ss << "]}";
+    return ss.str();
+  }
+};
+
+// Define values that might change depending on new segments added to
+// the planning buffer.
+// Currently planned profile for an AxisTarget.
+//    |  v1
+//    |   /¯¯¯¯¯¯¯¯¯¯\
```
hzeller, line 96: to avoid having trouble with the backslas at the end of the line, maybe add a comment at end
```
//    |  v1
//    |   /¯¯¯¯¯¯¯¯¯¯\    // trapezoidal curve.
//    |  /            \v2
//    | /
// v0 |/________________
//      acc  travel   dec
```

src/planner.cc
```diff
@@ -38,17 +47,77 @@ namespace {
 // The speed is initially the aimed goal; if it cannot be reached, the
 // AxisTarget will be modified to contain the actually reachable value. That is
 // used in planning along the path.
+// The starting_speed is the required speed at the start of the segment.
+// It is used to combine segments with different trajectory angles.
+// In the AxisTarget we define "motion invariants" (or targets). Values and
+// parameters that assuming the requested final trajectory stays the same, they
+// are constants, no matter the previous or next segments.
 struct AxisTarget {
   int position_steps[GCODE_NUM_AXES];  // Absolute position at end of segment.
```
hzeller, line 56: Suggest to use
```
AxesRegister position_steps;
AxesRegister delta_steps;
```
instead of he raw arrays.
(would require to change the memset with call to AxesRegister::zero() and other places that use a pointer to use a `const AxesRegister&` instead.

src/planner.cc
```diff
@@ -310,201 +381,145 @@ double Planner::Impl::euclidian_speed(const struct AxisTarget *t) {
...
+    }
 
-    // Now map axis steps to actual motor driver
+    // Travel
     for (const GCodeParserAxis a : AllAxes()) {
-      const int decel_steps = std::lround(decel_fraction * axis_steps[a]);
-      assign_steps_to_motors(&decel_command, a, decel_steps);
+      assign_steps_to_motors(&move_command, a,
+                             segment->target.delta_steps[a] -
+                               accel_command.steps[a] - decel_command.steps[a]);
```
hzeller, line 472: `accel_command` and `decel_command` are of type `LinearSegmentSteps`.
Their `steps` array only holds `BEAGLEG_NUM_MOTORS` elements, not the number covered by all the axes.
So here is a mix-up between axis and motor.

src/planner.cc
hzeller: (mmh, AxesRegister is float, so we need `FixedArray<int, GCODE_NUM_AXES, GCodeParserAxis>`.)

src/planner.cc
lromor: fixed by using `/* */`

src/planner.cc
```diff
@@ -38,17 +47,81 @@ namespace {
...
+  int delta_steps[GCODE_NUM_AXES];  // Difference to previous position. (steps)
   enum GCodeParserAxis defining_axis;  // index into defining axis.
-  double speed;       // (desired) speed in steps/s on defining axis.
-  uint16_t aux_bits;  // Auxillary bits in this segment; set with M42
-  double dx, dy, dz;  // 3D delta_steps in real units
-  double len;         // 3D length
+  double start_speed;  // Starting speed of the segment (steps/s).
+  double speed;        // Maximum speed of the defining axis (steps/s).
+  double accel;        // Maximum acceleration of the defining axis (steps/s^2).
+  unsigned short aux_bits;  // Auxillary bits in this segment; set with M42
```
hzeller, line 65: This is where the clang-tidy complains: this shall be `uint16_t` to be more explicit about the width (as `short` is not really be defined).

lromor: (solved)

src/planner.cc
lromor: nice catch! (solved)

src/planner.cc
lromor: (solved)

hzeller:
Very nice, the hard deceleration now is completely gone just looking at the `make test-html` output.

It does seem to accumulate some slight rounding errors along the way, most noticeable in this spiral cut example, in which the new machine path (colored) slightly deviates from the gcode path (solid thin line) at the end of the travel.

  Old  | New
-------|--------
spiral-cut-old | spiral-cut-new

Slightly easier to see in this direct comparison

spiral-cut-compare

I think we need some unit test that test this with
  * Very small segments (1 step long). Possibly also intermixed with 0 steps (which should not throw off the planner)
  * switching defining axis between segments. So say given `(Axis-X, Axis-Y)` we have a segment sequence of `(2,1), (1,2), (2,1), (1,2)...`  and possibly also a test with just one step like in the previous test case `(1,0), (0,1), (1,0), (0,1)...`

hzeller:
You might now need to re-base this pull request to incorporate the planner test changes.

src/planner.cc
```diff
@@ -81,20 +154,42 @@ static float clamp_defining_axis_limit(const int *axes_steps,
   return new_defining_axis_limit * steps_per_mm[defining_axis];
 }
 
+// Uniformly accelerated ramp distance estimation.
+// The ramp is parametrized by v0 as the starting speed, v1 the final speed
+// and the constant acceleration.
+static double uar_distance(double v0, double v1, double acceleration) {
```
hzeller, line 160: The uar is a needlessly cryptic prefix `uar_distance()`, `uar_speed()`...
If we can use longer names we should.

src/planner_test.cc
```diff
@@ -214,28 +246,23 @@ static void VerifySegmentConstraints(const MachineControlConfig &config,
...
+    const double factor = get_speed_factor_for_axis(&segment, i);
+    const double v1 = factor * segment.v0;
+    const double v2 = factor * segment.v1;
+    const double accel = fabs_accel(v1, v2, steps);
     const double max_accel =
       config.acceleration[axis] * config.steps_per_mm[axis];
     const double max_feedrate =
       config.max_feedrate[axis] * config.steps_per_mm[axis];
     EXPECT_LE(segment.v0, max_feedrate);
     EXPECT_LE(segment.v1, max_feedrate);
```
lromor, line 258: This doesn't look like a proper check. We should test (once) that the euclidean speed is indeed below the requested feedrate. Nothing else.

lromor:
Has you suggested, let's add another test checking the behavior for different planning buffer capacities.

hzeller (commented):
This looks very good!
Here a first round of comments.

src/planner.cc
```diff
@@ -557,21 +631,174 @@ bool Planner::Impl::machine_move(const AxesRegister &axis, float feedrate) {
...
+
+  // Flush the queue.
+  issue_motor_move_if_possible(true);
+}
+
+// Perform a backward and forward pass across
+// the whole planning buffer to adapt the start
+// and final speed of all the segments.
+void Planner::Impl::UpdateMotionProfile() {
+  assert(planning_buffer_.size());
```
hzeller, line 643: After a rebase, this can be `assert(!planning_buffer_.empty());`
(and clang-tidy will complain if it is not :) ).

src/planner.cc
```diff
@@ -310,201 +409,146 @@ double Planner::Impl::euclidian_speed(const struct AxisTarget *t) {
...
-#else
-  const char do_accel = 1;
-#endif
-
-  bool has_accel = false;
-  bool has_move = false;
-  bool has_decel = false;
+  // We require a slot to be always present.
+  if (num_segments == 0 &&
+      (planning_buffer_.size() == PLANNING_BUFFER_CAPACITY - 1))
```
hzeller, line 437: With the latest version in head, we have now also planning_buffer_.capacity(), so this can be formulated nicer.

(also note, that capacity() of the buffer is actually `PLANNING_BUFFER_CAPACITY - 1`, as the RingDeque needs one empty slot internally)

src/planner.cc
```diff
@@ -310,201 +409,146 @@ double Planner::Impl::euclidian_speed(const struct AxisTarget *t) {
...
+// motion
+//    that will always end up at 0 speed.
+// 4: Enqueue the front of the planning buffer to the motors if certain
+//    conditions occur.
 bool Planner::Impl::machine_move(const AxesRegister &axis, float feedrate) {
   assert(position_known_);  // call SetExternalPosition() after DirectDrive()
+
+  // We always assume there's enough space to store a new segment and that we
+  // have at least one segment.
+  assert((planning_buffer_.size() < PLANNING_BUFFER_CAPACITY - 1) &&
```
hzeller, line 540: With the version at head, there is now a `capacity()` call, so if you rebase, you can formulate that as
```
 assert(planning_buffer_.size() < planning_buffer_.capacity() && !planning_buffer_.empty());
```

hzeller:
With this change, the test with the square that is compared with single-segment lines vs. multi segment lint

* One-segment lines [square-moves-direct.gcode](https://github.com/hzeller/beagleg/blob/master/src/testdata/square-moves-direct.gcode)
* Multi segment lines [square-moves-small-segments.gcode](https://github.com/hzeller/beagleg/blob/master/src/testdata/square-moves-small-segments.gcode)

Is now properly accelerating and starting to decelerate in the middle. However, wouldn't we expect the speed profile now be exactly the same ? From the visualization it looks like that the small-segment version accelerates and decelerates quicker, leaving a longer stretch of fast movement ("yellow") in the middle. It looks like the postscript output has the same color/speed legend, so what is different ?

square-moves

src/planner.cc
lromor: (solved)

src/planner_test.cc
```diff
@@ -82,29 +83,34 @@ class FakeMotorOperations : public SegmentQueue {
...
+            "  Receiving: (%6.1f, %6.1f, %6.1f); Euclid space: "
             "len: %5.1f ; v: %8.1f -> %8.1f %s %s",
-            euclidian_speeds.steps[AXIS_X] / config_.steps_per_mm[AXIS_X],
-            euclidian_speeds.steps[AXIS_Y] / config_.steps_per_mm[AXIS_Y],
-            euclidian_speeds.steps[AXIS_Z] / config_.steps_per_mm[AXIS_Z],
-            hypotenuse,
-            segment.v0, segment.v1,
-            (segment.v0 < segment.v1) ? "accel" :
-            (segment.v0 > segment.v1) ? "decel" : "move",
+            (config_.steps_per_mm[AXIS_X] != 0)
```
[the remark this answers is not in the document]
lromor: (solved)

src/planner_test.cc
```diff
@@ -125,9 +131,15 @@ class FakeMotorOperations : public SegmentQueue {
   // Out of convenience, return back the length of the segment in euclid space
   float FixEuclidSpeed(LinearSegmentSteps *seg) {
     // Real world coordinates
-    const float dx = seg->steps[AXIS_X] / config_.steps_per_mm[AXIS_X];
-    const float dy = seg->steps[AXIS_Y] / config_.steps_per_mm[AXIS_Y];
-    const float dz = seg->steps[AXIS_Z] / config_.steps_per_mm[AXIS_Z];
+    const float dx = (config_.steps_per_mm[AXIS_X] != 0)
```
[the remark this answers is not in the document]
lromor: (solved)

src/planner_test.cc
```diff
@@ -417,30 +442,105 @@ TEST(PlannerTest, SimpleMove_AxisSpeedLimitClampsOverallSpeed_YY) {
...
+TEST(PlannerTest, ShortSegment_EdgeCases) {
+  PlannerHarness plantest(5.0f, 0);
+  MachineControlConfig &config = *plantest.GetConfig();
+  config.steps_per_mm[AXIS_X] = 1;
+  config.steps_per_mm[AXIS_Y] = 1;
+  config.acceleration[AXIS_X] = 99;
+  config.acceleration[AXIS_Y] = 99;
+
+  const float kFeedrate = 100.0f;
+  const int kDefiningAxisSwingSteps = 100;
```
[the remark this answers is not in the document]
lromor: I decided to remove the attribute, it's not important.

src/planner_test.cc
```diff
@@ -491,6 +591,116 @@ void testShallowAngleAllStartingPoints(float threshold, float testing_angle) {
...
+**    | /\    \
+** v0 |/__\____\_________
+**                       t
+*/
+TEST(PlannerTest, StraightLine_TwoSegmentsNeedsForwardPass) {}
+
+// Executing multiple short steps on a line should
+// should abide to the provided constraints.
+TEST(PlannerTest, StraightLine_LotsOfSteps) {
+  MachineControlConfig *config = new MachineControlConfig;
```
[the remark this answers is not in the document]
lromor: (solved)

src/planner.cc
```diff
@@ -310,201 +409,146 @@ double Planner::Impl::euclidian_speed(const struct AxisTarget *t) {
...
+      move_command.v1 = segment->planned.v1;
     }
-  }
 
-  // Move is everything that hasn't been covered in speed changes.
-  // So we start with all steps and subtract steps done in acceleration and
-  // deceleration.
-  for (const GCodeParserAxis a : AllAxes()) {
-    assign_steps_to_motors(&move_command, a, axis_steps[a]);
+    // Now we substract from travel both accel and travel.
```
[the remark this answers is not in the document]
lromor: (solved)

src/planner_test.cc
```diff
@@ -491,6 +591,116 @@ void testShallowAngleAllStartingPoints(float threshold, float testing_angle) {
...
+** enough that the previous segment needs a forward pass as well.
+** Currently planned profile for an AxisTarget.
+**    |___________
+**    |   /¯\
+**    |  /   \ v2
+**    | /\    \
+** v0 |/__\____\_________
+**                       t
+*/
+TEST(PlannerTest, StraightLine_TwoSegmentsNeedsForwardPass) {}
```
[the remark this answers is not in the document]
lromor: This is a duplicate concept. Removing it.

src/planner.cc
```diff
@@ -310,201 +409,146 @@ double Planner::Impl::euclidian_speed(const struct AxisTarget *t) {
...
+    subtract_steps(&move_command, accel_command);
+    const bool has_move = subtract_steps(&move_command, decel_command);
```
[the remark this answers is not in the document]
lromor: This test was made to check that: https://github.com/hzeller/beagleg/pull/44/files#diff-5d3dac6530af9bb210c33f63ea33d03de13fdeabf1bc36f74b93432805506c52R480

src/planner.cc
lromor: (solved)

src/planner.cc
lromor: (solved)

src/planner.cc
lromor: (solved)

src/planner.cc
```diff
@@ -310,201 +409,146 @@ double Planner::Impl::euclidian_speed(const struct AxisTarget *t) {
...
+//    that will always end up at 0 speed.
+// 4: Enqueue the front of the planning buffer to the motors if certain
+//    conditions occur.
 bool Planner::Impl::machine_move(const AxesRegister &axis, float feedrate) {
   assert(position_known_);  // call SetExternalPosition() after DirectDrive()
+
+  // We always assume there's enough space to store a new segment and that we
+  // have at least one segment.
+  assert((planning_buffer_.size() < PLANNING_BUFFER_CAPACITY - 1) &&
+         planning_buffer_.size());
```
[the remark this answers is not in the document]
lromor: (solved)

lromor:
> With this change, the test with the square that is compared with single-segment lines vs. multi segment lint
> 
> * One-segment lines [square-moves-direct.gcode](https://github.com/hzeller/beagleg/blob/master/src/testdata/square-moves-direct.gcode)
> * Multi segment lines [square-moves-small-segments.gcode](https://github.com/hzeller/beagleg/blob/master/src/testdata/square-moves-small-segments.gcode)
> 
> Is now properly accelerating and starting to decelerate in the middle. However, wouldn't we expect the speed profile now be exactly the same ? From the visualization it looks like that the small-segment version accelerates and decelerates quicker, leaving a longer stretch of fast movement ("yellow") in the middle. It looks like the postscript output has the same color/speed legend, so what is different ?
> 
> square-moves

I made a small python utility to compare the speed/steps plots of two gcode strings, these are the results for the two ways of doing a straight line:

image

They seem to nicely overlap. I'll try to dig deeper.

bigguiness:
Hi Leonardo,

I have been busy redesigning our larger motion control system and have not been able to actively follow your acceleration stuff.

Is this ready for testing on actual hardware or are you still working on the test code? I can give it a spin on my hardware when it’s ready and give you feedback on how it physically looks.

Regards,
Hartley

hzeller:
>  Is this ready for testing on actual hardware or are you still working on the test code?
> I can give it a spin on my hardware when it’s ready and give you feedback on how it physically looks.

Yes, the improved acceleration/deceleration code should be ready now.

Lines with many small segments (such as circles or arcs) that previously would result in an abrupt deceleration at the last segment, are now starting to decelerate early enough to meet deceleration constraints. I've run it on an actual machine and its moves and deceleration are smooth.

Code is slightly slower than before because of more calculations, but that only really shows in direct performance comparisons without a physical machine attached and shouldn't impact real-world speed (though we have it on the TODO list to see what can be improved).

The [install document](https://github.com/hzeller/beagleg/blob/main/INSTALL.md) also has been updated to describe how to install BeagleG on a current Beaglebone Debian image.

PR #47, opened 2021-09-11
Accel clamp

lromor:


src/planner.cc
```diff
@@ -52,6 +52,34 @@ struct AxisTarget {
...
+//
+// Example with speed: given i the axis with the highest (relative) out of
+// bounds speed, what's the speed of the defining_axis so that every speed
+// respects i's bounds? We just need to project each speed to the defining
+// axis speed, and pick the smallest.
+//
+// We need to work in mm, since steps is not a uniform unit of space across
+// the different axes.
+static float clamp_defining_axis_limit(const int *axes_steps,
+                                    const FloatAxisConfig &axes_limits_mm,
```
hzeller, line 68: This indentation looks off. I just checked, the clang-format was apparently an older version.
Can you re-base to head run `.github/bin/run-clang-format.sh` ?

src/planner_test.cc
```diff
@@ -17,6 +16,69 @@
 // TODO: This should work being set to 1
 #define SPEED_STEP_FACTOR 4
 
+// Helper function for implementing ASSERT_NEAR.
+testing::AssertionResult DoubleNearAbsRelPredFormat(const char* expr1,
```
hzeller, line 20: this looks more complicated than I'd expect, in particular since below in your test, you really only test relative errors, and have the absolute error set to zero.

So I'd reduce it: only write more code once it is needed.

src/planner_test.cc
```diff
@@ -250,13 +315,23 @@ static void parametrizedAxisClamping(GCodeParserAxis defining_axis,
                                      GCodeParserAxis slowAxis) {
   const float kClampFeedrate = 10.0;
   const float kFastFactor = 17;  // Faster axis is faster by this much.
+  const float kRelativeError = 0.1;
```
hzeller, line 318: a full 10% ?

hzeller (changes requested):
Nice. Looks like clamping is getting there.
A few comments below; I think in particular the new EXPECT is more complicated than needed and tests for two things at once. Just test for one thing (relative) and then if you need also absolute testing, then I'd suggest to just add another regular EXPECT call at the place where you need it. That makes it very clear what is tested and where things are failing.

src/planner.cc
```diff
@@ -52,6 +52,34 @@ struct AxisTarget {
...
+                                    enum GCodeParserAxis defining_axis,
+                                    const FloatAxisConfig &steps_per_mm) {
+  float ratio, new_defining_axis_limit = axes_limits_mm[defining_axis];
```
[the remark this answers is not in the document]
lromor: (solved)

src/planner_test.cc
lromor: (solved)

src/planner_test.cc
lromor: (solved)

src/planner.cc
lromor: (solved)

src/planner.cc
```diff
@@ -52,6 +52,37 @@ struct AxisTarget {
...
+static float clamp_defining_axis_limit(const int *axes_steps,
+                                       const FloatAxisConfig &axes_limits_mm,
+                                       enum GCodeParserAxis defining_axis,
+                                       const FloatAxisConfig &steps_per_mm) {
+  float new_defining_axis_limit = axes_limits_mm[defining_axis];
+  for (const GCodeParserAxis i : AllAxes()) {
+    const float ratio =
+      std::fabs((1.0 * axes_steps[i] * steps_per_mm[defining_axis]) /
+                (axes_steps[defining_axis] * steps_per_mm[i]));
+
```
hzeller, line 77: so looking at the line above, we  can only be zero if the numerator is zero. Given that we can assume that the defining axis has `steps_per_mm` defined we could bail out early with a `continue`, so that below we don't have do do the awkward zero check:

```c++
for(...) {
  if (axes_steps[i] == 0) continue;
  const float ratio = ...;
  new_defining_axis_limit = std::min(new_defining_axis_limit, axes_limits_mm[i] / ratio);
}
```

src/planner_test.cc
```diff
@@ -250,13 +274,24 @@ static void parametrizedAxisClamping(GCodeParserAxis defining_axis,
                                      GCodeParserAxis slowAxis) {
   const float kClampFeedrate = 10.0;
   const float kFastFactor = 17;  // Faster axis is faster by this much.
+  const float kRelativeError = 0.001;
```
hzeller, line 277: Maybe I'd call it `kAllowedRelativeError` so that it is more clear that we use that for testing.

src/planner_test.cc
```diff
@@ -17,6 +16,28 @@
...
+  return testing::AssertionFailure()
+         << "The difference between " << expr1 << " and " << expr2 << " is "
+         << diff << ", which exceeds " << rel_error_expr << " * max(abs("
+         << expr1 << "), abs(" << expr2 << ")), where\n"
+         << expr1 << " evaluates to " << val1 << ",\n"
+         << expr2 << " evaluates to " << val2 << ",\n"
+         << rel_error_expr << " evaluates to " << rel_error << ".";
+}
+
+#define EXPECT_NEAR_REL(val1, val2, abs_error, rel_error) \
```
hzeller, line 38: You don't use `abs_error` anymore, so that parameter should go :)

hzeller (approved):
Looks good, but a few small comments.

src/planner.cc
lromor: (solved)

src/planner_test.cc
lromor: (solved)

src/planner_test.cc
lromor: (solved)

hzeller:
now only one run of the formatter needed.

PR #53, opened 2022-02-13
Fixed previous check on speed continuity, added check max accel and f…

lromor:
…eedrate.

### DIFF
--- a/src/planner_test.cc
+++ b/src/planner_test.cc
@@ -38,6 +38,14 @@ testing::AssertionResult DoubleNearAbsRelPredFormat(const char *expr1,
 #define EXPECT_NEAR_REL(val1, val2, rel_error) \
   EXPECT_PRED_FORMAT3(DoubleNearAbsRelPredFormat, val1, val2, rel_error)
 
+// Compute the absolute acceleration given space s and
+// starting speed v0 and end speed v1.
+template <typename T>
+static inline T fabs_accel(const T v0, const T v1, const int s) {
+  const T accel = (v1 * v1 - v0 * v0) / (2 * s);
+  return fabs(accel);
+}
+
 // Set up config that they are the same for all the tests.
 static void InitTestConfig(struct MachineControlConfig *c) {
   float steps_per_mm = 1000;
@@ -186,6 +194,8 @@ class PlannerHarness {
     return motor_ops_.segments();
   }
 
+  MachineControlConfig *GetConfig() const { return config_; }
+
  private:
   MachineControlConfig *config_;
   FakeMotorOperations motor_ops_;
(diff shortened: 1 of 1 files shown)


src/planner_test.cc
```diff
@@ -194,9 +204,42 @@ class PlannerHarness {
...
+      const double accel =
+        fabs_accel(segment.v0, segment.v1, segment.steps[axis]);
+      const double max_accel =
+        config.acceleration[axis] * config.steps_per_mm[axis];
+      const double max_feedrate =
+        config.max_feedrate[axis] * config.steps_per_mm[axis];
+      EXPECT_LE(segment.v0, max_feedrate);
+      EXPECT_LE(segment.v1, max_feedrate);
+      if (segment.v0 != segment.v1)
+        EXPECT_NEAR_REL(accel, max_accel, rel_error);
```
lromor, line 225: Maybe let's test with `EXPECT_LE`?

src/planner_test.cc
```diff
@@ -194,9 +204,42 @@ class PlannerHarness {
...
+    }
+  }
+}
+
+size_t GetDefiningMotor(const LinearSegmentSteps &segment) {
```
hzeller, line 230: maybe make this `static`

src/planner_test.cc
```diff
@@ -209,9 +252,14 @@ static void VerifyCommonExpectations(
   // The joining speeds between segments match.
   for (size_t i = 0; i < segments.size() - 1; ++i) {
     // Let's determine the speed in euclidian space.
+    const size_t m1 = GetDefiningMotor(segments[i]);
+    const size_t m2 = GetDefiningMotor(segments[i + 1]);
+    if (m1 == m2) {
+      EXPECT_EQ(segments[i].v1, segments[i + 1].v0)
+        << "Joining speed between " << i << " and " << (i + 1);
+    }
 #if 1
```
hzeller, line 261: I guess we don't need the `#if 1` around here anymore, just leave it always on. It looks like it was `#if`-ed as there were reasons why it failed before.

hzeller (commented):
LGTM, two smallish remarks.

src/planner_test.cc
hzeller: Yes, as we only want to know that we never exceed the acceleration, but smaller is ok (in particular as we'll have jerk handling at some point, in which we expect it to be slower initially)

lromor: (solved)

src/planner_test.cc
lromor: (solved)

src/planner_test.cc
lromor: (solved)

PR #54, opened 2022-03-26
added Dynamic RingDeque container

lromor:


src/common/container.h
```diff
@@ -80,61 +81,99 @@ class FixedArray {
...
+  static_assert(CAPACITY > 1, "Capacity needs to be > 1");
+  // intentionally not initializing memory to better see if users do.
+  T buffer[CAPACITY];
+  using type = T;
+  T *GetBuffer() { return &buffer[0]; }
+  static constexpr size_t storage_size = CAPACITY;
+};
+
+template <typename T>
+struct DynamicStoragePolicy {
```
hzeller, line 100: 'Dynamic' is a bad name, as it is not dynamically re-allocating anything. It is still a fixed-size container, so the naming shtould reflect that.

src/common/container.h
```diff
@@ -80,61 +81,99 @@ class FixedArray {
...
 
-  size_t size() const { return (write_pos_ + CAPACITY - read_pos_) % CAPACITY; }
+  using type = T;
+  size_t storage_size;
+  std::unique_ptr<T[]> buffer;
+  T *GetBuffer() { return buffer.get(); }
+};
+
+template <typename StoragePolicy>
+class RingDeque : private StoragePolicy {
```
hzeller, line 113: For clarity, ideally, we should use composition, not inheritance.
I suppose you attempted it without, but then ran into the problem that it requires a constructor with size for the runtime container ?

hzeller (commented):
A first pass.

src/common/container.h
```diff
@@ -80,61 +81,99 @@ class FixedArray {
...
-  RingDeque() {
-    static_assert(CAPACITY > 1, "Capacity needs to be > 1");
-    // intentionally not initializing memory to better see if users do.
+template <typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args &&...args) {
+  return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+
+template <typename T, size_t CAPACITY>
+struct FixedStoragePolicy {
```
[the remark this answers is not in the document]
lromor: For now I'm settling for:
```c++
template <typename T, size_t CAPACITY>
using FixedRingDeque = RingDeque<internal::FixedArrayBuffer<T, CAPACITY>>;

// Identical to the FixedRingDeque but runtime-allocated.
template <typename T>
using RungimeAllocatedFixedRingDeque = RingDeque<internal::RuntimeAllocatedBuffer<T>>;
```

Let me know if you are satisfied!

src/common/container.h
lromor: (solved) See previous comment

src/common/container.h
lromor: I think it's not an issue to go for composition, it's just more boilerplate to add. With inheritance things are fairly more straightforward although a bit less flexible. Let me know your preference!

Issue #48, opened 2022-01-30
'/sys/devices/bone_capemgr.*/slots': No such file or directory


monomente:
I am trying to get started with the CRAMPS tree overlay.

When I run sudo` ./start-devicetree-overlay.sh CRAMPS/BeagleG-CRAMPS.dts`
I get the following output:
```
`make: 'CRAMPS/BeagleG-CRAMPS-00A0.dtbo' is up to date.

Adding BeagleG-CRAMPS overlay
./start-devicetree-overlay.sh: line 69: /sys/devices/bone_capemgr.*/slots: No such file or directory
cat: '/sys/devices/bone_capemgr.*/slots': No such file or directory`
```

It seems that the `bone_capemgr` directory is no longer present.  It seems that for newer versions the beaglebone might not compile and load the device tree overlays in this way.  I will try to read more to understand how this works but wondering if someone else has had this issue.

hzeller:
Yes, the Beaglebone has changed these things around a little in recent years; also the software interface to the PRU needs to be adapted.

I kept  the operating system version on my 'production' BeagleG CNC machines so things might be a bit 'crusty' depending on older Kernels.. In paritcular I found that at some point the whole software interface to the PRU (formly `uio_pruss`) just got more compilcated than needed so I didn't bother chasing the interface change.

Having said that, getting BeagleG working on a modern kernel on a Beaglebone would be highly appreciated, and we're ready to accept a pull request for that. Possibly the folks over at https://forum.beagleboard.org/ might be able to give a hand.
Jason Kridner @jadonk (say hi!) might know someone who can help converting old cape definitions and PRU PRUSS interface to whatever is current today.

(Currently I am working on some more platform independent version of BeagleG that will also run on other platforms).

jadonk:
Yeah, I'd love to answer a few questions to help this along!

hzeller:
Thanks for chiming in Jason!

As you know, BeagleG is a few years old now, and I have it running on my home CNC machines (and soon also my laser cutter), but the kernel interfaces to the hardware changed over time and while I was waiting for things to settle, I just pinned some older version on my Beagleboards. Bringing this up to a state that it is usable for a current operating system would be the goal.

There are mostly two toplevel questions (at this point):

[...]

PR #57, opened 2022-07-09
config-pin location has changed; provide path to script.

hzeller:
The old 'config-pin' script is not compatible with the `config-pin`
binary found in newer installations. Until the *.pin files are adapted,
make sure to have the documentation point to the right place.


hzeller:
First part for #48 : document the current location of the `config-pin` utility.

CC @bigguiness

PR #59, opened 2022-07-16
Improve planner performance

lromor:


src/planner.cc
```diff
@@ -394,6 +405,16 @@ Planner::Impl::Impl(const MachineControlConfig *config,
     const float accel = cfg_->acceleration[i] * cfg_->steps_per_mm[i];
     max_axis_accel_[i] = accel;
   }
+#if 0
```
hzeller, line 408: There is a trick with the comma to avoid looking for `size() - 1` every time to check if we need to add a comma  ... just check if we are the first element, otherwise prepend a comma; this makes it more compact:

```c++
#if 0
  fprintf(stderr, "\nmax_accelerations:\n[");
  for (unsigned i = 0; i < max_axis_accel_.size(); ++i) {
    fprintf(stderr, "%s%f", !i ? "" : ",", max_axis_accel_[(GCodeParserAxis)i]);
  }
  fprintf(stderr, "]\n");
#endif
```

(similar in other places in this file)

src/planner_test.cc
```diff
@@ -831,6 +831,40 @@ TEST(PlannerTest, DetectOutRangeAcceleration) {
...
+// at constat max feedrate we expect the profile to always be symmetrical.
+TEST(PlannerTest, StraightLine_ProfileSymmetry) {
+  MachineControlConfig *config = new MachineControlConfig();
+  InitTestConfig(config);
+  // The acceleration should stay high enough to avoid making
+  // to avoid numerical issues due to square root calculations
+  // for which the acceleration and deceleration, if they
+  // require too many steps, will end up being very inaccurate.
+  PlannerHarness plantest(0, 0, config);
+  unsigned kNumSteps = 200;
```
hzeller, line 844: Don't use `unsigned` unless really needed. Recommending `int`.
( https://google.github.io/styleguide/cppguide.html#Integer_Types )

hzeller, line 844: We're enquening kNumSteps segments, not steps. So wouldn't `kNumSegments` be a slightly better name ?
Also, since it is a constant in our function, should we make it `const` ?

src/planner_test.cc
```diff
@@ -831,6 +831,40 @@ TEST(PlannerTest, DetectOutRangeAcceleration) {
...
+  AxesRegister pos = {};
+
+  for (unsigned i = 0; i < kNumSteps; ++i) {
+    pos[AXIS_X] += 10;
+    plantest.Enqueue(pos, 10000);
+  }
+  // Check the profile is symmetric.
+  const auto segments = plantest.segments();
+  int32_t accel_steps = 0;
+  int32_t decel_steps = 0;
```
hzeller, line 854: Maybe we should also record total sum of steps and then see if we get `kNumSegments * kStepsPerSegment` in total just as a test that we don't loose a step.

src/planner_test.cc
```diff
@@ -831,6 +831,40 @@ TEST(PlannerTest, DetectOutRangeAcceleration) {
...
+  // The acceleration should stay high enough to avoid making
+  // to avoid numerical issues due to square root calculations
+  // for which the acceleration and deceleration, if they
+  // require too many steps, will end up being very inaccurate.
+  PlannerHarness plantest(0, 0, config);
+  unsigned kNumSteps = 200;
+  AxesRegister pos = {};
+
+  for (unsigned i = 0; i < kNumSteps; ++i) {
+    pos[AXIS_X] += 10;
```
hzeller, line 848: If we have a constant `kStepsPerSegment`, this number doesn't come out of thin air (and it would allow to change it for experiments)

src/planner_test.cc
```diff
@@ -831,6 +831,40 @@ TEST(PlannerTest, DetectOutRangeAcceleration) {
...
+    plantest.Enqueue(pos, 10000);
```
hzeller, line 849: 10000 also looks like a constant that might be good if named.

hzeller (commented):
Nice, this is a good check that the planner does the right thing.

src/planner_test.cc
lromor: (solved)

lromor: (solved): true

src/planner_test.cc
lromor: (solved)

src/planner_test.cc
lromor: (solved)

src/planner_test.cc
lromor: (solved)

src/planner.cc
```diff
@@ -659,7 +681,7 @@ void Planner::Impl::UpdateMotionProfile() {
 
   // Perform the backward pass. Compute and join the travel and deceleration
   // parts of the profile.
-  for (; buffer_index < planning_buffer_.size(); --buffer_index) {
+  for (; buffer_index >= last_planned_invariant_segment_; --buffer_index) {
```
hzeller, line 684: For semicolon-separated parts that are not needed in a `for`-loop, I usually add a `/**/`, just so that it is more visible.

```c++
for (/**/; buffer_index >= last_planned_invariant_segment_; --buffer_index) {
```

Not needed now, but could be considered in the future.

hzeller (approved):
Nice!

PR #61, opened 2022-08-05
Test and fix for moving non-euclidean axes in arcs

lucalenardi:
We are using beagleg (thanks for this super nice project, btw!) to drive stepper motors systems for industrial machines. In most cases, the system is made by XYZ axis, plus one or more axes to drive single or multiple extruders (A, B, E etc). Those tools needs to be synced with the motion during the process, that's why interpolating A, B... axes with X, Y, Z is actually needed.
 
It works perfectly for points and lines, but not with arcs, where non-euclidean axes are ignored.

This pull request tries to solve this issue by adding a test (with an arc on the XY plane and a linear motion for A) and a small fix in `arc_gen.cc` to make it pass.

### DIFF
--- a/src/gcode-parser/arc-gen_test.cc
+++ b/src/gcode-parser/arc-gen_test.cc
@@ -59,6 +59,7 @@ class TestArcAccumulator : public GCodeParser::EventReceiver {
   }
 
   float total_len() const { return total_len_; }
+  AxesRegister last() const { return last_; }
 
   bool rapid_move(float feed_mm_p_sec, const AxesRegister &absolute_pos) final {
     return true;
@@ -153,6 +154,28 @@ static void testArcLength(GCodeParserAxis normal, bool clockwise, double start,
     EXPECT_NEAR(collect.total_len(), expected_len, 0.003);
   }
 }
+
+TEST(ArcGenerator, ArcLength_NonEuclideanAxes) {
+  AxesRegister start, center, target;
+  const double kTurnAngle = M_PI;
+  const float kAxisALenght = 20.0;
+
+  start[AXIS_X] = 1.0;
+  start[AXIS_Y] = 0.0;
+  start[AXIS_A] = 0.0;
+
+  center[AXIS_X] = 0;
+  center[AXIS_Y] = 0;
+
+  target[AXIS_X] = cos(kTurnAngle);
+  target[AXIS_Y] = sin(kTurnAngle);
+  target[AXIS_A] = kAxisALenght;
+
+  TestArcAccumulator collect(start);
+  collect.arc_move(100, AXIS_Z, true, start, center, target);
+  EXPECT_NEAR(collect.last()[AXIS_A], kAxisALenght, 0.003);
+}
+
 
 TEST(ArcGenerator, ArcLength_CW_360) {
   testArcLength(AXIS_X, true, TEST_ARC_STEP, TEST_ARC_STEP, 2 * M_PI);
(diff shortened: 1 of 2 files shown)


src/gcode-parser/arc-gen_test.cc
```diff
@@ -154,6 +155,28 @@ static void testArcLength(GCodeParserAxis normal, bool clockwise, double start,
   }
 }
 
+TEST(ArcGenerator, ArcLength_NonEuclideanAxes) {
```
hzeller, line 158: To make more clear what we're testing, I'd probably name it
`ArcLength_InterpolateNonEuclidianAxes`.

src/gcode-parser/arc-gen_test.cc
```diff
@@ -154,6 +155,28 @@ static void testArcLength(GCodeParserAxis normal, bool clockwise, double start,
...
+  AxesRegister start, center, target;
+  const double kTurnAngle = M_PI;
+  const float kAxisALenght = 20.0;
```
hzeller, line 161: Lenght : typo, last two letters swapped.

src/gcode-parser/arc-gen_test.cc
```diff
@@ -59,6 +59,7 @@ class TestArcAccumulator : public GCodeParser::EventReceiver {
   }
 
   float total_len() const { return total_len_; }
+  AxesRegister last() const { return last_; }
```
hzeller, line 62: I'd make that a `const AxesRegister&`  reference.

hzeller:
How do you define the additional axis in the `G2` or `G3` call ?

I guess you can just give it an `A` position as target position, but I think that is never tested anywhere.
So I suggest to also add a test for that in [gcode-parser_test.cc](https://github.com/hzeller/beagleg/blob/4693e81c0271b76292bcb8455824e01ee0f7ba50/src/gcode-parser/gcode-parser_test.cc#L600) so that this feature is documented.

hzeller:
I suppose the axes other than X, Y and Z would also be good to interpolate for the `G5` and `G5.1` spline features. I am pretty sure it has the same problem that you fixed for the circles. If you want to have a follow-up pull request :)

hzeller:
to fix the formatting problem the continuous integration complains about, just run

```
.github/bin/run-clang-format.sh
```

hzeller:
Alright, I checked: the gcode parser was also not taking the additional axis into account. That is now fixed in head ([this was the change](https://github.com/hzeller/beagleg/commit/b377e316473f9c5fb4014ba0532e484886fc4965) ... including some unrelated clean-up and improved tests).

I have added a [section in the test](https://github.com/hzeller/beagleg/blob/bec5bf057b63bc553acdbf237c704215a458889d/src/gcode-parser/gcode-parser_test.cc#L700-L704) that now fails (so this is why it is `#if`-def'ed out), but should pass after your change is in.

lucalenardi:
> I suppose the axes other than X, Y and Z would also be good to interpolate for the `G5` and `G5.1` spline features. I am pretty sure it has the same problem that you fixed for the circles. If you want to have a follow-up pull request :)

Yes definitely something I will look at.

hzeller (approved):
LGTM

hzeller:
Thanks! Merged.

You now have the honor of removing the `#if 0` around here, as it will now be working
https://github.com/hzeller/beagleg/blob/1cf62595bf927bc221f991adc547cbe8eeac329f/src/gcode-parser/gcode-parser_test.cc#L700-L704

PR #62, opened 2022-12-02
Planner refactoring

lromor:


src/planner.cc
```diff
@@ -663,175 +668,205 @@ void Planner::Impl::bring_path_to_halt() {
   issue_motor_move_if_possible(true);
 }
 
+// Compute the backard pass of <segment_to_plan> given the
+// information (such as starting speed and defining axis) of the <next_segment>.
+// If <next_segment> is nullptr, the next segment is assumed to not exist (end
+// of the buffer). The function returns true if the final speed already connects
+// with the starting speed of the next segment and this is not the last segment.
```
hzeller, line 675: The return value might be slightly confusing. The function is named `BackwardUpdateSegment()` so one would expect that `true` actually means that it was updated.
Might be worthwhile reconsidering the return value.

src/planner.cc
```diff
@@ -663,175 +668,205 @@ void Planner::Impl::bring_path_to_halt() {
   issue_motor_move_if_possible(true);
 }
 
+// Compute the backard pass of <segment_to_plan> given the
```
hzeller, line 671: The word _Backward_ and _Forward_ might make sense in the way we compute them when looking through our queue, but that is also just an artifact how we organized our queue.
Maybe we should name these functions in the functionality they do, e.g. function names such as `DecelerationPlanSegment()` and `AccelerationPlanSegments()` ? Then when reading the loop itself, it then also emphasises to the reader which directions our queue is set up.
Then there is less ambiguity as looking at the names of the functions explains

src/planner.cc
```diff
@@ -663,175 +668,205 @@ void Planner::Impl::bring_path_to_halt() {
...
+  // Final speed constraint.
+  const double end_speed = (speed_factor != 0) ? next_v0 / speed_factor : 0;
+
+  // Check if the current segment v2 speed connects with the next
+  // (except if this is the very last segment).
+  // If we do, then we stop here. No need to compute the accel. It will happen
+  // in the forward pass. The new profile deceleration and travel will always
+  // be above or equal to the current profile. This will only then happen if
+  // the new backward profile v2 equals the previous planned profile v2.
+  if (planned.v2 == end_speed && !is_last_segment) {
```
hzeller, line 718: Sometimes comparing _exactly_ with float numbers is doable in particular if we have assigned things from the same calculation, but I think it makes sense to comment _why_ comparing two floats exactly is safe here
(as opposed to comparing that it is in very close proximity or some relative e.g. `<=` operator).

src/planner.cc
```diff
@@ -663,175 +668,205 @@ void Planner::Impl::bring_path_to_halt() {
...
   // since we plan to always decelerate to zero.
-  double speed = 0;
-  GCodeParserAxis defining_axis = planning_buffer_.back()->target.defining_axis;
+  PlanningSegment *next_segment = nullptr;
+  PlanningSegment *previous_segment = nullptr;
 
-  // Perform the backward pass. Compute and join the travel and deceleration
-  // parts of the profile.
+  // Perform a backward pass.
   for (; buffer_index >= last_planned_invariant_segment_; --buffer_index) {
```
hzeller, line 847: For a follow-up PR it might be worthwhile to explore if we can just iterate through the buffer (I think currently the container does not have an iterator yet, so that would need to be added). For later.

src/planner.cc
```diff
@@ -663,175 +668,205 @@ void Planner::Impl::bring_path_to_halt() {
...
-      segment->planned.v0 = segment_start_speed;
-      segment->planned.accel = steps_to_vmax;
-      segment->planned.travel = segment->planned.travel - steps_to_vmax;
-      speed = segment->planned.v2;
-      continue;
+    next_segment = planning_buffer_[buffer_index];
+    if (ForwardUpdateSegment(previous_segment, next_segment) && !first_accel) {
+      // We have acceleration.
+      // Let's update the last invariant segment.
+      last_planned_invariant_segment_ = buffer_index;
```
hzeller, line 866: The name `last_planned_invariant_segment` is not really conveying much information for the reader. We should brainstorm for a better name. It should highlight that this is the cut-off point for a fully-planned output.

(also the question: when it is fully planned, could we use that opportunity to flush the buffer to the underlying hardware at this point ?)

hzeller (commented):
Nice!
Mostly just considerations for naming and meaning of return values so far.

src/planner.cc
lromor: (solved)

src/planner.cc
lromor: (solved)

src/planner.cc
lromor: (solved)

src/planner.cc
lromor: (solved): using num_segments_ready.

hzeller:
LGTM

PR #66, opened 2023-04-16
Added command M181.

lromor:
- Parser shall route to unprocessed commands.
- GCodeMachineControl shall handle the command and set the value to max if no S<value> pair is provided. Added single test that only checks the motor ops queue shall be emptied before changing the planner size.
- Planner respect the lookahead value and always push older commands to maintain planning queue size. No tests added.

### DIFF
--- a/src/gcode-machine-control.cc
+++ b/src/gcode-machine-control.cc
@@ -122,6 +122,7 @@ class GCodeMachineControl::Impl final : public GCodeParser::EventReceiver {
                     int max_steps);
   void home_axis(enum GCodeParserAxis axis);
   void set_output_flags(HardwareMapping::NamedOutput out, bool is_on);
+  void set_lookahead(int size);
   void handle_M105();
   // Parse GCode spindle M3/M4 block.
   const char *set_spindle_on(bool is_ccw, const char *);
@@ -450,6 +451,22 @@ bool GCodeMachineControl::Impl::check_for_pause() {
   return hardware_mapping_->TestPauseSwitch();
 }
 
+void GCodeMachineControl::Impl::set_lookahead(int size) {
+  // It is necessary to have a fully stopped machine to change lookahead.
+  // This is to avoid introducing a delay between the requested change and
+  // actual machine behavior with previously enqueued segments.
+  motor_ops_->WaitQueueEmpty();
+  const bool set_max = size < 0;
+  size = (size <= 0) ? -1 : size;
+  const bool set = planner_->SetLookAhead(&size);
+  if (set_max) {
+    planner_->SetLookAhead(&size);
+  }
+  if (!set) {
+    mprintf("// ERROR: allowed range 1 <= S <= %d\n", size);
+  }
+}
+
 void GCodeMachineControl::Impl::handle_M105() {
   mprintf("// ");
   for (int chan = 0; chan < 8; chan++) {
@@ -534,6 +551,14 @@ const char *GCodeMachineControl::Impl::special_commands(char letter,
   case 119: mprint_endstop_status(); break;
   case 120: pause_enabled_ = true; break;
   case 121: pause_enabled_ = false; break;
+  case 181: {
+    const char *const after_pair = parser_->ParsePair(remaining, &letter, &value, msg_stream_);
+    if (remaining == NULL) set_lookahead(-1);
+    else if (letter == 'S') set_lookahead((unsigned)value);
+    else break;
+    remaining = after_pair;
+    break;
+  }
   default:
     mprintf("// BeagleG: didn't understand ('%c', %d, '%s')\n",
             letter, code, remaining);
--- a/src/gcode-machine-control.cc
+++ b/src/gcode-machine-control.cc
@@ -75,6 +75,9 @@ class GCodeMachineControl::Impl final : public GCodeParser::EventReceiver {
   HomingState GetHomeStatus();
   bool GetMotorsEnabled();
   void GetCurrentPosition(AxesRegister *pos);
+  bool SetLookahead(int size) { return planner_->SetLookahead(size); }
+  int Lookahead() const { return planner_->Lookahead(); }
+  int GetMaxLookahead() const { return planner_->GetMaxLookahead(); }
 
   // -- GCodeParser::Events interface implementation --
   void gcode_start(GCodeParser *parser) final;
@@ -122,7 +125,7 @@ class GCodeMachineControl::Impl final : public GCodeParser::EventReceiver {
                     int max_steps);
   void home_axis(enum GCodeParserAxis axis);
   void set_output_flags(HardwareMapping::NamedOutput out, bool is_on);
-  void set_lookahead(int size);
+  const char *handle_set_lookahead(const char *);
   void handle_M105();
   // Parse GCode spindle M3/M4 block.
   const char *set_spindle_on(bool is_ccw, const char *);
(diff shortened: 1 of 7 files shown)


src/planner.h
```diff
@@ -64,6 +64,16 @@ class Planner {
...
 
+  // Set/get internal planning buffer size. Trade-off between
+  // command sent - executed motion latency and maximum achievable speed.
+  // Set the internal planning queue to the value pointed by size.
+  // If size is nullptr, this is a no-op and false is returned.
+  // Special size 0 returns false and sets size to the current value.
+  // Negative size values, or size values above the internal maximum
+  // will not change the internal state and set size to the maximum value
+  // instead.
+  bool SetLookAhead(int *size);
```
hzeller, line 75: Is t here a good reason to make an unusually complicated multi-purpose function to set and get ?

I'd just apply the KISS principle and have three functions that do one thing and everyone understands intuitively:

```c++
 bool SetLookahead(int size);
 int Lookahead() const;
 int GetMaxLookahead() const;
```

src/planner.cc
```diff
@@ -934,6 +938,23 @@ void Planner::Impl::SetExternalPosition(GCodeParserAxis axis, float pos) {
...
+bool Planner::Impl::SetLookAhead(int *size) {
+  if (size == nullptr) {
+    return false;
+  }
+  const int value = *size;
+  if (value == 0) {
+    *size = lookahead_size_;
+    return false;
+  }
+  if (value < 0 || value > PLANNING_BUFFER_CAPACITY) {
```
hzeller, line 950: This allows to set the planning buffer to zero. Would that work, or do we need at least one slot ?

src/planner.cc
```diff
@@ -277,6 +278,9 @@ class Planner::Impl {
   // segments and update only the last deceleration ramp.
   int num_segments_ready_ = 0;
 
+  // Number of maximum planning steps allow to enqueue.
+  int lookahead_size_ = PLANNING_BUFFER_CAPACITY;
```
hzeller, line 282: I'd make this `size_t` so that you don't have to use the awkward unsigned cast in Line 470.
This value is vetted by the `SetLookahead()` function already to make sure it never is negative.

src/gcode-machine-control.cc
```diff
@@ -450,6 +451,22 @@ bool GCodeMachineControl::Impl::check_for_pause() {
...
+  // actual machine behavior with previously enqueued segments.
+  motor_ops_->WaitQueueEmpty();
+  const bool set_max = size < 0;
+  size = (size <= 0) ? -1 : size;
+  const bool set = planner_->SetLookAhead(&size);
+  if (set_max) {
+    planner_->SetLookAhead(&size);
+  }
+  if (!set) {
+    mprintf("// ERROR: allowed range 1 <= S <= %d\n", size);
```
hzeller, line 466: This is ok for now, I think in the future we need to re-think `gcode_block_done()` that it also returns an error status. So nothing to do in this PR.

src/gcode-machine-control.cc
```diff
@@ -534,6 +551,14 @@ const char *GCodeMachineControl::Impl::special_commands(char letter,
   case 119: mprint_endstop_status(); break;
   case 120: pause_enabled_ = true; break;
   case 121: pause_enabled_ = false; break;
+  case 181: {
+    const char *const after_pair = parser_->ParsePair(remaining, &letter, &value, msg_stream_);
+    if (remaining == NULL) set_lookahead(-1);
+    else if (letter == 'S') set_lookahead((unsigned)value);
```
hzeller, line 557: If you're casting the value here to unsigned already, you might be in for a surprise: if someone sets `M181 S-1`, then you're setting values to int-max here. But ideally we want set set look ahead function have a chance to complain that it is below or above the value. Now it is always above the value.

src/gcode-machine-control.cc
```diff
@@ -534,6 +551,14 @@ const char *GCodeMachineControl::Impl::special_commands(char letter,
   case 119: mprint_endstop_status(); break;
   case 120: pause_enabled_ = true; break;
   case 121: pause_enabled_ = false; break;
+  case 181: {
+    const char *const after_pair = parser_->ParsePair(remaining, &letter, &value, msg_stream_);
```
hzeller, line 555: I suggest to move the whole logic in a separate `handle_set_lookahead()` function.
(see `set_spindle_on()` as an example).

hzeller (commented):
First round of comments.

src/gcode-machine-control_test.cc
```diff
@@ -219,6 +227,25 @@ TEST(GCodeMachineControlTest, straight_segments_speed_change) {
...
+  };
+  Harness harness(expected);
+  GCodeParser::Config config;
+  GCodeParser::Config::ParamMap parameters;
+  config.parameters = &parameters;
+  GCodeParser parser = GCodeParser(config, harness.gcode_emit());
+  EXPECT_EQ(0, parser.error_count());
+
+  harness.gcode_emit()->gcode_start(&parser);
+  parser.ParseBlock("M181 S100", nullptr);
```
hzeller, line 242: Also add tests for `M181` without setting the S value (as it then should go back to planner size) and with negative values as well as too large values to see how things behave.

src/gcode-machine-control_test.cc
```diff
@@ -110,8 +118,8 @@ class Harness {
     return machine_control->ParseEventReceiver();
   }
 
-  MockMotorOps expect_motor_ops_;
-  HardwareMapping hardware_;
+  MockMotorOps expect_motor_ops;
+  HardwareMapping hardware;
```
hzeller, line 122: we never access hardware, so we should be explicit and make it private (and keep the underscore), and put `machine_control` and `expect_motor_ops` in the public.

src/planner.h
```diff
@@ -64,6 +64,18 @@ class Planner {
   // Precondition: BringPathToHalt() had been called before.
   void SetExternalPosition(GCodeParserAxis axis, float pos);
 
+  // Set the internal planner queue size.
+  // Used to trade-off between command sent - executed motion latency
+  // and maximum achievable speed.
+  // Requires 0 < size <= GetMaxLookahead(), returns true on success.
+  bool SetLookahead(int size);
```
hzeller, line 71: Now that the planner has some way to set the queue size: are there tests we want in the planner_test.cc as well ? e.g. figure out how the acceleration properties change when we change the queue size ?

src/gcode-machine-control_test.cc
```diff
@@ -35,6 +35,9 @@ static void init_test_config(struct MachineControlConfig *c,
 }
 
 namespace {
+// Record Segment queue calls we care avout.
+enum { CALL_wait_queue_empty, NUM_COUNTED_CALLS };
```
hzeller, line 39: I am not entirely sure if it needs so complicated with enum, a Count() function and a publically accessible array.
Couldn't that just be a simple int ?

(it looks like this is prepared for 'future use'. But we should decide that when the future is there and we need 10 different counters).

src/gcode-machine-control_test.cc
lromor: (solved)

src/gcode-machine-control_test.cc
lromor: (done)

src/gcode-machine-control_test.cc
lromor: (solved)

lromor (commented):
Added the requested changes. Now working on a test for the planner.

lromor:
@hzeller , I've added a planner test to check the lookahead resize behavior. Let me know what you think!

hzeller (approved):
A line in the [G-Code documentation](https://github.com/hzeller/beagleg/blob/main/G-code.md#m-codes-dealt-with-by-gcode-machine-control) would probably a good follow-up change.

hzeller:
In the `gcode_finished()` callback the GCodeMachineControl receives: might be useful to also reset the planner buffer to normal.

PR #67, opened 2023-04-23
M181 - docs

lromor:


G-code.md
```diff
@@ -159,6 +159,7 @@ M117             | Display message.
 M119             | Get endstop status.
 M120             | Enable pause switch detection.
 M121             | Disable pause switch detection.
+M181 Snn         | Set planning buffer size to nn. Requires the machine to stop.
```
hzeller, line 162: _'Requires the machine to stop'_ sounds like that is a requirement that the user needs to fulfill.
_'Will flush the current queue first'_  would probably better explain what is happening.

Also, add a description that without parameter, it sets it to the default size.

PR #68, opened 2023-04-29
Queue clearance for emergency stop

lromor:


src/motion-queue.h
```diff
@@ -118,6 +118,10 @@ class MotionQueue {
   // The return parameter head_item_progress is set to the number
   // of not yet executed loops in the item currenly being executed.
   virtual int GetPendingElements(uint32_t *head_item_progress) = 0;
+
+  // Perform an immediate reset of the queue,
+  // even if motors are still moving.
+  virtual bool Clear() = 0;
```
hzeller, line 124: The 'even if motors are still moving': does that mean best-effort to stop motors ASAP ? Then this should be mentioned here (Imagine someone implementing it: they read the comment and need to come up with an implementation that matches that expectation).
What does the return value mean ?

src/motion-queue-motor-operations_test.cc
```diff
@@ -202,6 +211,35 @@ TEST(RealtimePosition, zero_loops_edge) {
...
+// The physical status should be reset and motion_backend.Clear() called.
+TEST(RealtimePosition, clear_queue) {
+  HardwareMapping hw;
+  MockMotionQueue motion_backend = MockMotionQueue();
+  MotionQueueMotorOperations motor_operations(&hw, &motion_backend);
+
+  // Enqueue a segment
+  LinearSegmentSteps segment = {
+    0 /* v0 */,
+    0 /* v1 */,
```
hzeller, line 224: If both speeds are zero, how would we move ?

src/pru-motion-queue.cc
```diff
@@ -177,6 +177,11 @@ void PRUMotionQueue::Shutdown(bool flush_queue) {
   MotorEnable(false);
 }
 
+bool PRUMotionQueue::Clear() {
+  pru_interface_->Shutdown();
```
hzeller, line 181: this sounds a bit drastic. Is that all we can do ?

hzeller (commented):
first round of comments.

src/segment-queue.h
```diff
@@ -77,6 +77,12 @@ class SegmentQueue {
   // source (e.g. homing). This will allow accurate reporting of the
   // PhysicalStatus.
   virtual void SetExternalPosition(int axis, int position_steps) = 0;
+
+  // Clear the queue and reset it to its initial state.
+  // This action is immediate and will discard any running or
+  // enqueued and not yet executed segment.
+  // Current physical status will be lost.
```
hzeller, line 84: What does the return value mean ?

src/motion-queue-motor-operations_test.cc
lromor: (solved)

src/segment-queue.h
lromor: (solved)

src/motion-queue-motor-operations_test.cc
```diff
@@ -202,6 +205,32 @@ TEST(RealtimePosition, zero_loops_edge) {
...
+    CreateMockSegment({10, 20, 30, 40, 50, 60, 70, 80});
+  int expected[BEAGLEG_NUM_MOTORS] = {10, 20, 30, 40, 50, 60, 70, 80};
+
+  motor_operations.Enqueue(segment);
+  motion_backend.SimRun(0, 1);
+
+  PhysicalStatus status;
+  motor_operations.GetPhysicalStatus(&status);
+  EXPECT_THAT(expected, ::testing::ContainerEq(status.pos_steps));
+  motor_operations.HaltAndDiscard();
```
hzeller, line 227: To keep the readers' train of thought, I'd group the  things with newlines to better see the before/call/after situation:

```c++
  PhysicalStatus status;
  motor_operations.GetPhysicalStatus(&status);
  EXPECT_THAT(expected, ::testing::ContainerEq(status.pos_steps));

  motor_operations.HaltAndDiscard();
  EXPECT_EQ(motion_backend.clear_calls_count, 1);

  memset(expected, 0, sizeof(expected));
  motor_operations.GetPhysicalStatus(&status);
  EXPECT_THAT(expected, ::testing::ContainerEq(status.pos_steps));
```

hzeller (commented):
LGTM.

Did you verify that stoping and starting will work on the PRU, and that you can restart and start enquing things again that come out on the other oscilloscope end ?

src/motion-queue-motor-operations_test.cc
lromor: (solved)

PR #80, opened 2026-06-27
Add flake.nix wrapping shell.nix

lromor:
## Summary
- Adds a `flake.nix` whose `devShells.default` imports the existing `shell.nix`, so `nix develop` works alongside `nix-shell` with a single source of truth for dependencies.
- No `flake.lock` is committed: inputs float on `nixos-unstable`, matching the existing `shell.nix` behaviour (which already tracks `<nixpkgs>`). Nix will auto-create a local lock on first run.

## Test plan
- [x] `nix-shell --run "clang-format --version"` still works.
- [x] `nix develop --command bash -c "clang-format --version"` enters the same environment.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,14 @@
+{
+  description = "BeagleG dev shell";
+
+  # nixos-unstable @ 2026-06-26
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/e73de5be04e0eff4190a1432b946d469c794e7b4";
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,13 @@
+{
+  description = "BeagleG dev shell";
+
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/4062d36ebeae843c750011eef6b61ec9a9dbc9a9";
```
hzeller: Maybe add a comment from when this is, like nixpkgs stable xx.xx on date yyyy-dd-mm

hzeller (approved):
LGTM

PR #88, opened 2026-07-30
Add PRU backend for stock mainline-kernel images: PRU_BACKEND=genirq.

lromor:
Recent BeagleBone Debian images ship neither uio_pruss nor PRU rpmsg support, leaving BeagleG without a way to run there. The genirq backend needs no PRU kernel driver: the host resets the core, copies the pasm-assembled firmware into PRU0 IRAM (with a readback check: IRAM writes are silently dropped on a core that is not freshly reset) and enables it, all through /dev/mem. Per-segment completion interrupts arrive through the kernel uio_pdrv_genirq driver, wired to PRU system event 19 by a small device-tree overlay with the same event->channel->host INTC mapping that prussdrv programs for the uio backend.

Firmware and default build are unchanged: same motor-interface-pru.p, same pasm flow; the uio_pruss backend stays the default.

Verified on a factory BeagleBone Black Rev D with the stock Debian image: ~5 kHz step pulses on P9_18 matching the planned segment timing, and the interrupt path measured end-to-end (the blocking read wakes ~1 ms after segment completion).


src/pru-hardware-interface.h
hzeller: Since there are now two, we should probably describe this one as well and distinguish what is used in what context (kernel version, kernel flavor (TI or mainline..)

src/pru-hardware-interface.h
```diff
@@ -53,4 +54,28 @@ class UioPrussInterface : public PruHardwareInterface {
...
+// PRU hardware control without any PRU-specific kernel driver, for
+// stock mainline-kernel BeagleBone images: firmware load, start/stop
+// and the shared motion queue through the memory maps of a generic UIO
+// device, and per-segment completion interrupts through its blocking
+// read, all wired up by the device-tree overlay in
+// dts/BEAGLEG-PRU-IRQ.dts (see INSTALL.md).
+class UioGenirqInterface : public PruHardwareInterface {
+ public:
+  UioGenirqInterface();
+  ~UioGenirqInterface() final;
```
hzeller, line 72: since you supply a destructor that is final, also make the class intheritance final (some compilers might be picky).
While at it, do that also at the `UioPrussInterface`.

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,174 @@
...
+
+#include "common/logging.h"
+#include "pru-hardware-interface.h"
+
+// Generated PRU code from motor-interface-pru.p
+#include "motor-interface-pru_bin.h"
+
+namespace {
+// Sizes of the UIO mmap regions; their physical addresses and order
+// are defined by the reg entries in dts/BEAGLEG-PRU-IRQ.dts.
```
hzeller: citation needed: since we're using this here with specific constants for the PRU and amx335x, we should have a link to a datasheet section or source code that explain these magic values.
(also below kCtrlEnable and such)

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,174 @@
...
+constexpr int kCtrlMap = 1;
+constexpr size_t kPru0CtrlSize = 4096;
+constexpr int kIramMap = 2;
+constexpr size_t kPru0IramSize = 8 * 1024;
+
+// PRU_CTRL: bit 0 is SOFT_RST_N (writing 0 resets the core), bit 1 is
+// ENABLE. "Enable" keeps SOFT_RST_N low, so the core starts from
+// instruction 0.
+constexpr uint32_t kCtrlReset = 0;
+constexpr uint32_t kCtrlEnable = 2;
```
hzeller: If this is indeed bits not enum values (datasheet reference needed), this might be more readable as 
  kCtrlEnable = (1 << 1) and kCtrlHalt = (1 << 0)

dts/BEAGLEG-PRU-IRQ.dts
```diff
@@ -0,0 +1,31 @@
...
+ * backend), plus the PRU0 memories as UIO mmap regions so no /dev/mem
+ * access is needed. The reg order is the backend's mmap contract:
+ * map 0 = PRU0 data RAM, map 1 = PRU0 control, map 2 = PRU0 IRAM.
+ * Needs uio_pdrv_genirq.of_id=generic-uio on the kernel command line.
+ */
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "ti,am335x-bone-black", "ti,am335x-bone", "ti,am33xx";
```
hzeller: beaglebone green is also a commonly used board. Is it compatible with that ?

dts/BEAGLEG-PRU-IRQ.dts
```diff
@@ -0,0 +1,31 @@
+/* BeagleG: expose the PRU to userspace as /dev/uioN via uio_pdrv_genirq:
+ * PRU system event 19 (motion-segment completion) routed through the
+ * PRUSS interrupt controller as event 19 -> INTC channel 2 -> host
+ * interrupt 2 (the same mapping prussdrv programs for the uio_pruss
+ * backend), plus the PRU0 memories as UIO mmap regions so no /dev/mem
+ * access is needed. The reg order is the backend's mmap contract:
+ * map 0 = PRU0 data RAM, map 1 = PRU0 control, map 2 = PRU0 IRAM.
+ * Needs uio_pdrv_genirq.of_id=generic-uio on the kernel command line.
+ */
```
hzeller, line 33: is there a datasheet we can reference that explains the magic values and addresses ?

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,174 @@
...
+    return -1;
+  }
+  const int fd = open(dev, O_RDWR);
+  if (fd < 0) Log_error("open(%s): %s", dev, strerror(errno));
+  return fd;
+}
+}  // namespace
+
+UioGenirqInterface::UioGenirqInterface()
+    : ctrl_(nullptr), iram_(nullptr), dram_(nullptr), uio_fd_(-1) {}
```
hzeller: the members are in sequence ctrl_, iram_, dram_ ...

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,174 @@
...
+  if (dram_) munmap(dram_, kPru0DramSize);
+  if (iram_) munmap(iram_, kPru0IramSize);
+  if (ctrl_) munmap((void *)ctrl_, kPru0CtrlSize);
+  if (uio_fd_ >= 0) close(uio_fd_);
+}
+
+bool UioGenirqInterface::Init() {
+  uio_fd_ = OpenUioDevice();
+  if (uio_fd_ < 0) return false;
+  dram_ = MapUioRegion(uio_fd_, kDramMap, kPru0DramSize);
```
hzeller: but here, we're initializing them in order dram, ctrl_ and iram_.
Unless there is a good reason to do this in this order (which means: it requires a comment), we should initialize them in the order they are as members.

src/pru-hardware-interface.h
```diff
@@ -53,4 +54,28 @@ class UioPrussInterface : public PruHardwareInterface {
...
+
+  bool Init() final;
+  bool AllocateSharedMem(void **pru_mmap, size_t size) final;
+  bool StartExecution() final;
+  unsigned WaitEvent() final;
+  bool Shutdown() final;
+
+ private:
+  volatile uint32_t *ctrl_;  // PRU0 control register.
+  void *iram_;
```
hzeller: if these are indeed shared memory areas between PRU and host, I wonder if we need to consider these volatile ?

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,174 @@
...
+#include "pru-hardware-interface.h"
+
+// Generated PRU code from motor-interface-pru.p
+#include "motor-interface-pru_bin.h"
+
+namespace {
+// Sizes of the UIO mmap regions; their physical addresses and order
+// are defined by the reg entries in dts/BEAGLEG-PRU-IRQ.dts.
+constexpr int kDramMap = 0;
+constexpr size_t kPru0DramSize = 8 * 1024;
```
hzeller: given that we're talking about page sizes, maybe 2 * 4096 is a bit more readable ?

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,174 @@
...
+constexpr uint32_t kCtrlHalt = 1;
+
+// Our node name in dts/BEAGLEG-PRU-IRQ.dts; the kernel-reported UIO
+// device name may carry an @unit-address suffix, so we prefix-match.
+constexpr const char *kUioName = "beagleg_pru_irq";
+
+// UIO map N is selected by mmap-ing at offset N * page-size.
+void *MapUioRegion(int uio_fd, int map_index, size_t size) {
+  void *m = mmap(nullptr, size, PROT_READ | PROT_WRITE, MAP_SHARED, uio_fd,
+                 (off_t)map_index * sysconf(_SC_PAGESIZE));
```
hzeller: is the _SC_PAGESIZE 4k ? Because then the map indices and the sizes they map would overlap at least for dram and ctrl
  * dram: 0...8191 
  * ctrl: 4096...8191
  * iram: 8192...16384
  
 Is this intentional, or am I am missing something ?

src/uio-genirq-interface.cc
hzeller: instead of dram, maybe we should call it dataram ? Otherwise it sounds like dynamic RAM, which it hopefully isn't

hzeller (commented):
Looks promising. If possible, we need more links to documentation, so that it doesn't look like magic. In particular the ranges of dataram, instructionram and ctrl seem to be interesting as they seem to overlap ?

src/uio-genirq-interface.cc
```diff
@@ -103,50 +110,51 @@ int OpenUioDevice() {
...
 
 UioGenirqInterface::UioGenirqInterface()
-    : ctrl_(nullptr), iram_(nullptr), dram_(nullptr), uio_fd_(-1) {}
+    : uio_fd_(-1), data_ram_(nullptr), ctrl_(nullptr), iram_(nullptr) {}
 
 UioGenirqInterface::~UioGenirqInterface() {
-  if (dram_) munmap(dram_, kPru0DramSize);
-  if (iram_) munmap(iram_, kPru0IramSize);
-  if (ctrl_) munmap((void *)ctrl_, kPru0CtrlSize);
+  if (data_ram_) munmap(data_ram_, kDataRamSize);
```
hzeller: in the destrucor, we should unmap them in reverse order (doesn't make a difference, but better follows the expected RAII patterns)

src/uio-genirq-interface.cc
```diff
@@ -44,30 +44,37 @@
...
 // Our node name in dts/BEAGLEG-PRU-IRQ.dts; the kernel-reported UIO
 // device name may carry an @unit-address suffix, so we prefix-match.
 constexpr const char *kUioName = "beagleg_pru_irq";
 
-// UIO map N is selected by mmap-ing at offset N * page-size.
+// The mmap offset on a UIO device is not a byte offset into one flat
+// space; it selects which memory region to map, and the kernel maps
+// that region from its start: "To map the memory of mapping N, you
+// have to use N times the page size as your offset"
+// (linux/Documentation/driver-api/uio-howto.rst).
```
hzeller: ok, so if we map ctrl memory one block in (4096 offset) from the uio_fd we get a a _different_ memory than the second half of the dataram that starts at position zero and goes up to 8192 ?

dts/BEAGLEG-PRU-IRQ.dts
```diff
@@ -52,4 +52,17 @@
 			};
 		};
 	};
+
+	/* BeagleG owns PRU0, so keep the kernel's pru_rproc from binding
```
hzeller, line 67: does this mean we need to load the device overlay in a particular oder ?

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,183 @@
...
+// Generated PRU code from motor-interface-pru.p
+#include "motor-interface-pru_bin.h"
+
+namespace {
+// The three UIO mmap regions, in the order the reg entries of
+// dts/BEAGLEG-PRU-IRQ.dts declare them. Their physical addresses live
+// in that overlay; region sizes per the AM335x Technical Reference
+// Manual (TI document SPRUH73, https://www.ti.com/lit/pdf/spruh73),
+// section 4.3 "PRU_ICSS Memory Map". Linux on the AM335x uses 4 KiB
+// pages.
```
hzeller: 4.3.2 shows the offsets at 0x0000_0000 for data ram, 0x0002_2000 for control and 0x0003_4000 for PRU0 iram.
which is the global mapping, but it does not describe the memory map regions for UIO

lromor:
I reviewed code and validated the documentation links.

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,190 @@
...
+
+namespace {
+// The three UIO mmap regions, in the order the reg entries of
+// dts/BEAGLEG-PRU-IRQ.dts declare them. Their physical addresses live
+// in that overlay; region sizes per the AM335x Technical Reference
+// Manual (TI document SPRUH73), section 4.3.2, Table 4-8 "Global
+// Memory Map": https://www.ti.com/lit/pdf/spruh73#page=207 (anchor
+// per revision Q). Linux on the AM335x uses 4 KiB pages.
+constexpr size_t kPageSize = 4096;
+constexpr int kDataRamMap = 0;  // PRU0 data RAM, 8 KiB.
```
hzeller: It would probably make it more maintainable if we keep the relevant values together, e.g.  if we had a struct, e.g. something like

```c++
struct MapRange {
   size_t offset;  // or use map index as you see fit
   size_t size;
};
constexpr MapRange kDataBlock { .offset = 0 * kPageSize, .size = 2 * kPageSize };
constexpr MapRange kCtrlBlock { .offset = 1 * kPageSize, .size =  kPageSize };
constexpr MapRange kInstructionRamBlock {.offset = 2 * kPageSize, size = 2 * kPageSize };
```

And then later have the `MapUioRegion()` get a MapRange parameter.

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// Find and open the /dev/uioN backed by our device-tree node.
+int OpenUioDevice() {
+  // Buffers sized for the longest possible directory entry (NAME_MAX).
+  char dev[sizeof("/dev/") + NAME_MAX] = "";
+  DIR *dir = opendir("/sys/class/uio");
+  struct dirent *de;
+  while (dir && !dev[0] && (de = readdir(dir)) != nullptr) {
+    char path[sizeof("/sys/class/uio/") + NAME_MAX + sizeof("/name")];
```
hzeller, line 108: wouldn't we want to filter out "." and ".." ?

src/uio-genirq-interface.cc
```diff
@@ -0,0 +1,190 @@
...
+  memset(data_ram_, 0x00, size);
+  *pru_mmap = data_ram_;
+  return true;
+}
+
+bool UioGenirqInterface::StartExecution() {
+  // A PRU core that is not freshly reset silently drops IRAM writes, so
+  // reset first and read the code back to turn that failure mode into a
+  // hard error.
+  *ctrl_ = 0;  // SOFT_RST_N low: hold the core in reset.
```
hzeller: Wouldnt we mess with the other bits in the control register if we set everything to zero ?
I would've expected come classic like *ctrl_ &= ...
Or are the other bits in that register always zero, so we can do this ? (in that case: comment needed)

lromor:
Thanks! Addressed the comments. I would like to give it another round of tests though on the real hardware.

INSTALL.md
```diff
@@ -48,6 +48,60 @@ look like this:
 uboot_overlay_pru=/lib/firmware/AM335X-PRU-UIO-00A0.dtbo
 ```
 
+### Alternative: images without uio_pruss (PRU_BACKEND=genirq)
```
hzeller, line 51: Maybe in a follow up change, we should reverse the default: describe this as the 'standard way' and have the 'old way' described in the paragraph afterwards.

hzeller (approved):
Thanks for the investigation and thoroughly documented implementation.

